Issue #4, opened 2017-01-03
Documentation errors


dethe:
There are some cut-and-paste errors that result in some APIs for Methods or Organizations looking like duplicates of Case APIs and being grouped erroneously.

dethe:
Fixed in PR #5

PR #5, opened 2017-01-03
fix cut-and-paste errors

dethe:
Addresses #4. I had to use `npm run postinstall` to see the changes locally, but the generated files appear to be checked in, so I *think* a `git pull` will show the fixes.


dethe:
Given that the CI is failing no matter what, I think this code is ready for review. @davidascher can you take a look? It *should* just fix some errors in documentation.

PR #14, opened 2017-02-16
Adding SQL and CSV files (users faked)

dethe:
This is the core data, I still need to add bookmarks back in and hook the backend API up to it.


dethe:
@davidascher Ready for review, will make a new PR for the next bits of API replacement.

api/controllers/case.js
```diff
@@ -34,35 +35,39 @@ var jsonStringify = require('json-pretty');
...
-    for (let i in buckets) {
-      countryCounts[buckets[i].key] = buckets[i].doc_count
-    }
-    res.status(200).json({
-      OK: true,
-      data: {
-        countryCounts: countryCounts
-      }
+    db.query(
+        'select $1~.$3~, count($1~.$3~) from $1~, $2~ where $1~.$4~ = $2~.$5~ group by $1~.$3~;',
```
davidascher: Not now, but I'd hope there's a way of writing these queries in such a way that we use variable names that mean something to the reader, rather than have to do the mapping of indices to names, no?

api/controllers/case.js
```diff
@@ -93,16 +98,15 @@ router.post('/new', function (req, res, next) {
...
-      type: 'case',		
-      body: req.body		
-    }, function (error, response) {		
-      if (error) {		
-        res.status(error.status).json({message: error.message})		
-      } else {		
-        // console.log(response)		
-        res.status(200).json(req.body)		
-      }		
+      index: 'pp',
```
davidascher: Is the idea to do another pass w/ new API conversions later?  This will index in elasticsearch, but not in SQL, so any tests of this (not that we have any) would not work.

migrations/method__localized_texts.csv
```diff
@@ -0,0 +1,149 @@
+body,language,method_id,title
+"<p><a id=""Participants"" name=""Participan",en,145,21st Century Town Meeting
```
davidascher: Participan?

davidascher: omg is this really how CSV escapes "'s?

api/controllers/case.js
dethe: Yes, there is. I worked from an example first, but will be evolving these queries pretty quickly and using dict mapping rather than array mapping.

api/controllers/case.js
dethe: Yes, more API conversions to come as fast as possible. Ideally, with accompanying tests.

migrations/method__localized_texts.csv
dethe: I *think* that is carrying over a typo from localized body text. Or else something has gone horribly worng. I will investigate further.

dethe: Python CSV, anything is possible.

PR #19, opened 2017-02-28
18 case subtables

dethe:
Flesh out Case data. Found some issues with the case tables that I'll need to resolve in a separate issue.


api/controllers/case.js
```diff
@@ -191,7 +195,24 @@ router.put('/:caseId', function editCaseById (req, res) {
...
+                the_case.files.push(att);
+            }else if (att.type == 'image'){
+                if (att.is_lead){
+                    the_case.lead_image = att;
+                }else{
+                    the_case.other_images.push(att);
+                }
+            }
+        });
+        the_case.methods = data[3]
```
davidascher: This is another place where I don't like the fact that we end up having to match the order of SQL statements in the `batch` block with the order of unpacking the returned data in the `then()` block.  Don't know if there's better syntax for that, but it feels brittle in the face of changing code.

test/cases.js
```diff
@@ -65,4 +65,18 @@ describe('Cases', () => {
...
+    if('This case should have 3 tags', (done) => {
+      chai.request(app)
+        .get('/case/39')
+        .set('Content-Type', 'application/json')
+        .set('Accept', 'application/json')
+        .end((err, res) => {
+            res.body.OK.should.be.equal.to('true')
+            res.should.have.status(200);
+            let the_case = res.body.data;
+            the_case.should.have.lengthOf(3);
```
davidascher: This doesn't read as though it has three tags.  It reads as though somehow the "case" has length 3.  Which feels weeeird.

api/controllers/case.js
dethe: I agree, but in this case I don't have a good answer. I'd like to get more of the structure built directly in the SQL query if I can, but need to up my SQL game for that.

davidascher: Understood.  I don't have a better answer either, just flagging the risk.

test/cases.js
dethe: I think that's a typo, not sure why it passed. It should be `the_case.tags.should.have.lengthOf(3)` Now I'm worried that it passed.

davidascher: It didn't run this test: https://travis-ci.org/participedia/api/builds/206444406

davidascher: Because you don't have a `done()` call I believe

dethe: Ah, fair enough. Mocha automatically runs all tests in `/tests/`, right?

davidascher: I think so, but travis' log which would show the test actually being run isn't available because of a travis problem.

PR #21, opened 2017-03-02
20 user api

dethe:
Added API for users, with start of tests. While I was in there I also ported the GET methods for method and organization to use Postgres and added tests for them. I'll need to do some front-end updates before we can safely deploy this, but it is ready for review.


test/organizations.js
```diff
@@ -0,0 +1,67 @@
...
+        .set('Accept', 'application/json')
+        .send({
+        })
+        .end((err, res) => {
+          res.should.have.status(401);
+          done();
+        });
+    });
+    it('works with authentication', (done) => {
+      chai.request(app)
```
davidascher: Seems like we shouldn't allow the creation of an org w/ no body.

dethe: We can add a test for that.

PR #29, opened 2017-03-06
Lots of linting fixes, and some more rigorous approaches to style.

davidascher:
Prettier for formatting
Airbnb linting rules


davidascher:
Dethe, would appreciate you looking over this.  Most of it is mechanical.  I'll flag with comments some of the specific points.

api/controllers/bookmark.js
```diff
@@ -32,32 +31,32 @@ var log = require('winston')
...
-          status: 'success',
-          data: data,
-          message: 'Retrieved ALL users'
+          status: "success",
+          data,
+          message: "Retrieved ALL users"
         });
       })
-      .catch(function (err) {
+      .catch(error => {
```
davidascher: Note: this was probably a real bug.

api/controllers/bookmark.js
```diff
@@ -154,35 +167,39 @@ router.post('/add', function addBookmark (req, res, next) {
...
+            res.status(401).json({
+              message: "access denied - user is not the owner of the bookmark"
+            });
+          } else {
+            db
+              .none("delete from bookmarks where id = $1", bookmarkID)
+              .then(data => res.status(200).json({
+                status: "success",
+                message: 'Removed a bookmark',
+                data: data
```
davidascher: I'm passing the data back just in case.

api/controllers/user.js
```diff
@@ -119,50 +117,47 @@ router.get('/get/:userId', function edituserById (req, res, next) {
...
-        db.none('insert into users(name)' +
-            'values(${name})',
-          req.body.user)
-          .then(function () {
-            res.status(200)
-              .json({
-                status: 'success',
-                message: 'Inserted user'
-              });
+      .catch(function(err) { /* this feels like it assumes what the error was */
```
davidascher: This feels like an assumption in the code that I worry could mask other error types.  IIUC, you're assuming that the `catch` happens because there is no result to the `one` call?

api/helpers/groups.js
```diff
@@ -1,24 +1,28 @@
...
-        }
+    const user = req.user;
+    if (
+      user &&
+      user.app_metadata &&
+      user.app_metadata.authorization &&
+      user.app_metadata.authorization.groups
+    ) {
+      if (user.app_metadata.authorization.groups.indexOf(groupName) === -1) {
+        errCB && errCB();
```
davidascher: the previous `errCB(err)` would probably never happen as `err` was undefined.

PR #34, opened 2017-03-21
22 implement search

dethe:
OK, first the caveats:

* Needs to be ruggedized and tested more thoroughly
* Needs to be simplified (the SQL is *very* repetitive)
* Full-text search is not searching authors names yet
* Probably some of the single-choice text fields will need to go back to lists

But the SQL-ified database is at least as capable as the old document database now. Full-text search and all.


api/controllers/search.js
```diff
@@ -1,44 +1,184 @@
...
+                .json({ results: [{ type: objType, hits: objList }] });
+        })
+        .catch(function(error) {
+            log.error("Exception in GET /search/getAllForType", error);
+            res.status(500).json({ error: error });
+        });
+}
+
+function get_all_nouns(res, facets, page, language, orderBy) {
+    // IMPLEMENT ME!
```
davidascher: old comment?

dethe: Oopsie, yes.

dethe:
Yeah, hard to do a line-by-line check on "let's change everything" but now that it's merged we can create issues for anything that crops up.

PR #39, opened 2017-03-29
Auto-add users in user table as needed

davidascher:
For discussion.


api/controllers/case.js
```diff
@@ -147,40 +149,39 @@ router.post("/new", function(req, res, next) {
...
-            OK: true,
-            data: case_id
-          });
+      let user_id = req.user && req.user.user_id;
+      if (! user_id) {
+        user_id = getUserIdForUser(req.user, function(user_id) {
+          console.log("Create new case for %s", req.body);
+          db
+            .none(
+              sql("../sql/create_case.sql"),
```
davidascher: For reasons I don't understand, this fails with:

```
Executing query WITH insert_case as ( INSERT into cases ( original_language, issue, communication_mode, communication_with_audience, content_country, decision_method, end_date, facetoface_online_or_both, facilitated, voting, number_of_meeting_days, ongoing, post_date, published, start_date, total_number_of_participants, updated_date, targeted_participant_demographic, kind_of_influence, targeted_participants_public_role, targeted_audience, participant_selection, specific_topic, staff_type, type_of_funding_entity, typical_implementing_entity, typical_sponsoring_entity, who_else_supported_the_initiative, who_was_primarily_responsible_for_organizing_the_initiative, location, lead_image, other_images, files, videos, tags, featured ) VALUES ( 'en', null, null, null, null, null, null, null, null, 'none', null, false, 'now', true, 'now', null, 'now', 'General Public', null, 'Lay Public', 'General Public', 'Open to all', null, null, null, null, null, null, null, null, CAST(ROW('''', '', 0) as attachment), '{}', '{}', '{}', '{}', false ) RETURNING id as case_id ), insert_author as ( INSERT into case__authors(user_id, timestamp, case_id) VALUES (2, 'now', (select case_id from insert_case)) ) INSERT INTO case__localized_texts(body, title, language, case_id) VALUES ( 'sausec', 'tomatoasasd', 'en', (select case_id from insert_case) ) ;
error: Exception in POST /case/new => error: duplicate key value violates unique constraint "cases_pkey"
```

davidascher: fixed by #175

api/controllers/case.js
```diff
@@ -127,7 +128,20 @@ router.get("/countsByCountry", function(req, res) {
...
+    Object.keys(req.user).forEach(function(key) {
+      console.log("res.user.%s => %s", key, req.user[key]);
+    });
+    if (req.user.metadata) {
+      Object.keys(req.user.metadata).forEach(function(key) {
+        console.log("res.user.metadata.%s => %s", key, req.user.metadata[key]);
+      });
+    }
+  } else {
+    console.log(">> No user at all <<");
```
davidascher: note, this shouldn't happen because of `groups.user_has(...)`

api/sql/create_user_id.sql
```diff
@@ -0,0 +1,14 @@
...
+  email,
+  language,
+  language_1,
+  name
+)
+VALUES
+  (
+    ${userEmail},
+    'en', 'en',
+    'jon do'
```
davidascher: Note: fix this based on name from auth0.

api/controllers/case.js
dethe: Yes, that was an old debugging statement, just removed it.

api/sql/create_user_id.sql
dethe: Fixed in commit  ce4a92a

api/helpers/user.js
```diff
@@ -1,8 +1,11 @@
 var { db, sql } = require("../helpers/db");
 var log = require("winston");
 
-var getUserIdForUser = function (user, cb) {
-  log.info("user", user)
+var getUserIdForUser = function(user, cb) {
+  log.info("user", JSON.stringify(user));
```
davidascher: make optional or remove

api/helpers/user.js
```diff
@@ -6,8 +6,13 @@ function ensureUser(req, res, next) {
   let user = req.user;
   let name = req.header("X-Auth0-Name");
   let auth0UserId = req.header("X-Auth0-UserId");
+  if (!user) {
+    // not authenticated, let it fail elsewhere
```
davidascher: why not fail right here?

app.js
```diff
@@ -55,6 +56,7 @@ app.use(
     algorithms: ["HS256"]
   }).unless({ method: ["OPTIONS", "GET"] })
 );
+app.use(ensureUser.unless({ method: ["OPTIONS", "GET"] }));
```
davidascher: hmm, I wonder if technically we shouldn't exclude `HEAD` as well.

davidascher:
Looks good!

PR #48, opened 2017-04-04
35 complete add apis

dethe:
Adds new object based on data from QuickSubmit.

Requires that at least title and summary fields are filled in. All other fields are optional.
User is added to authors table for the new object
Related objects are added to their own (join) tables (ignoring duplicates)
Title and body (i.e., summary) are added to the localized_text table for the object type

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -212,7 +212,9 @@ router.post("/new", function(req, res, next) {
   let location = as.location(req.body.location);
   let videos = as.videos(req.body.vidURL);
   let lead_image = as.attachment(req.body.lead_image); // frontend isn't sending this yet
-  let related_cases = ""; // frontend isn't sending this yet
+  let related_cases = as.related_list(req.body.related_cases);
+  let related_methods = as.related_list(req.body.related_methods);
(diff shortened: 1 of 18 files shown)


api/controllers/case.js
```diff
@@ -226,6 +238,7 @@ router.post("/new", function(req, res, next) {
     })
     .catch(function(error) {
       log.error("Exception in POST /case/new => %s", error);
+      console.error(error);
```
davidascher: Why do we need both log.error and console.error?

api/helpers/db.js
```diff
@@ -20,16 +21,113 @@ try {
...
 }
 
-module.exports = {
-  db: db,
-  sql: sql,
-  as: pgp.as
-};
+// as.author
+function author(user_id, name) {
+  // TODO: escape user_id and name to avoid injection attacks
```
davidascher: Please file a P1 bug on this so we don't lose track.

api/controllers/case.js
dethe: Oh, one can be removed. I added the console error because I wasn't seeing the log when running tests that errored out, but I didn't see the console message either.

PR #50, opened 2017-04-05
Bookmark tweaks

davidascher:


api/controllers/bookmark.js
```diff
@@ -29,33 +31,42 @@ let log = require("winston");
...
+      });
+      log.error(error);
+    });
+}
+
+function queryBookmarks(req, res, next) {
+  let userid = req.params.userid;
+
+  if (!userid) {
+    ensureUser(req, res, function(req, res, next) {
```
dethe: We can add ensureUser as middleware on this route and avoid the test and duplication here.

dethe: Nevermind, I guess the idea here is that you can look at the bookmarks of any user, not just your own?

api/controllers/bookmark.js
```diff
@@ -83,59 +94,56 @@ router.get("/list/:userId", function(req, res, next) {
  */
 
 router.post("/add", function addBookmark(req, res, next) {
-  try {
-    if (!req.body.bookmarkType) {
-      log.error("Required parameter (bookmarkType) wasn't specified");
-      res.status(400).json({
-        message: "Required parameter (bookmarkType) wasn't specified"
+  if (!req.body.bookmarkType) {
```
dethe: We should validate and escape bookmarkType and thingId (userId too, but it is relatively safe after passing through our authentication process.

api/controllers/bookmark.js
```diff
@@ -154,32 +162,44 @@ router.post("/add", function addBookmark(req, res, next) {
  *
  */
 
-router.delete("/delete/:bookmarkID", function updateUser(req, res, next) {
+router.delete("/delete", function updateUser(req, res, next) {
   let userId = req.user.user_id;
-  let bookmarkID = parseInt(req.params.bookmarkID);
+  let bookmarktype = req.body.bookmarkType;
+  let thingid = req.body.thingID;
```
dethe: More SQL validation and escaping needed, but I can tackle that on the branch where I'm securing SQL

setup.sql
```diff
@@ -225,7 +225,7 @@ CREATE TABLE bookmarks (
   ID SERIAL PRIMARY KEY,
   bookmarktype VARCHAR,
   thingid INTEGER,
-  userid VARCHAR
+  userid INTEGER REFERENCES users(id)
```
dethe: Good fix!

api/controllers/case.js
```diff
@@ -305,16 +307,42 @@ router.put("/:caseId", function editCaseById(req, res) {
...
+      getUserIfExists(req, res, function(userId) {
+        if (userId) {
+          db
+            .one(
+              "select * from bookmarks where bookmarktype = $1 AND thingid = $2 AND userid = $3",
+              ["case", req.params.caseId, userId]
+            )
+            .then(function(data) {
+              console.info("CASE IS BOOKMARKED");
+              the_case["bookmarked"] = true;
```
dethe: This would be good to move into `[type]_by_id.sql`, but I can take care of that later.

PR #56, opened 2017-04-10
Script to extract the values of the various fields & l10n keys

davidascher:
@dethe --

Thinking that we can use something like this to extract the l10n keys and create the schema for the selection.  Not sure how we go about changing the CSV import code to do the same substitution.

Note the key function mapping value in english to the l10n nick is:

```
          let nick = value
            .replace("&amp", "")
            .replace("#039;", "")
            .replace(/\s+/g, "_")
            .replace(/[\.,\-()&\$£]/g, "")
            .toLowerCase();
```


davidascher:
Yes, this list is definitely one that could be cleaned up especially with someone who can authoritatively say "we'll throw this data away".  Note that I'm using the `nickify` function and the generated files in a couple of ways in the frontend:

- the `nickify` function I'm using to create nicks from the values returned by the API, "pretending" that I'm living in a future where the database has gone through the nickification process.  I'll remove that function when that happens, but it relies on the nickification being an algorithmic process.
- the JSON file is driving the `SelectField` widgets
- the locale data is driving the nickname->english mapping.

So when we go to bespoke mapping, it's important that we still generate these two files.

davidascher:
Still, it's starting to work:

screenshot 2017-04-09 20 31 15

(no idea where 'or deliberation' comes from yet)

davidascher:
(it's in the SQL db as it is, worth looking into)

PR #76, opened 2017-04-18
52 related things

### LINKED ISSUE #52: Case/method/org APIs don't return related things
- [ ] `/case/<num>` should return `related_cases`, `related_methods`, `related_orgs` in the payload
- [ ] same for `/method/<num>`
- [ ] same for `/organization/<num>`

dethe:
Fixes #52.

Also improves error reporting when querying for an object that doesn't exist (methods and organizations, had already fixed for cases).

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -312,7 +312,7 @@ async function getCaseById(req, res) {
     the_case.bookmarked = bookmarked.case;
     res.status(200).json({ OK: true, data: the_case });
   } catch (error) {
-    log.error("Exception in GET /case/undefined => %s", error);
+    log.error("Exception in GET /case/%s => %s", res.params.caseId, error);
     res.status(500).json({
       OK: false,
       error: error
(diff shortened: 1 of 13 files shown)


api/controllers/case.js
```diff
@@ -297,6 +297,7 @@ router.put("/:caseId", function editCaseById(req, res) {
 async function getCaseById(req, res) {
   try {
     const caseId = as.number(req.params.caseId);
+    console.log("caseId: %s", caseId);
```
davidascher: remove

api/controllers/case.js
```diff
@@ -311,7 +312,7 @@ async function getCaseById(req, res) {
     the_case.bookmarked = bookmarked.case;
     res.status(200).json({ OK: true, data: the_case });
   } catch (error) {
-    log.error("Exception in GET /case/%s => %s", caseId, error);
+    log.error("Exception in GET /case/undefined => %s", error);
```
davidascher: undefined?  Seems like the `catch` handler could be invoked even if `caseId` is defined.

PR #88, opened 2017-04-23
Harden userlookups

davidascher:


api/helpers/user.js
```diff
@@ -5,10 +5,14 @@ let unless = require("express-unless");
 const getUserIfExists = async req => {
   let user = req.user;
   if (user) {
-    user = await db.one(sql("../sql/user_by_email.sql"), {
+    user = await db.oneOrNone(sql("../sql/user_by_email.sql"), {
```
dethe: Weird, I thought we'd made this change awhile ago. Looks reasonable in any case.

api/controllers/case.js
```diff
@@ -303,12 +303,16 @@ async function getCaseById(req, res) {
...
     });
     const userId = await getUserIfExists(req);
-    const bookmarked = await db.one(sql("../sql/bookmarked.sql"), {
-      type: "case",
-      thingId: caseId,
-      userId: userId
-    });
-    the_case.bookmarked = bookmarked.case;
+    if (userId) {
+      const bookmarked = await db.one(sql("../sql/bookmarked.sql"), {
```
dethe: Do we need to add this in methods and organizations as well?

davidascher: Probably!

PR #101, opened 2017-05-02
parses whitespace now on server-side, dafaults to AND search

### LINKED ISSUE #83: Fix or document how to do multi-word searches
`/search?query=participatory%20budgeting` results in a nasty error message.

Need to know how to send full text queries that are multiple words.

dethe:
Replaced `to_tsquery` with `plainto_tsquery` which turns whitespace into `&` between non-stopwords. This is an AND search, returning results which include both (or all) words, but not relying on word order. If we want to retain word order, we can use `phraseto_tsquery` which replaces whitespace with the FOLLOWED BY operator.

We can get more fancy, supporting AND, OR, quotation-grouping, and FOLLOWED BY, but it will require us to handle more of the parsing on our side (or exposing Postgres operators `&`, `|`, `>>` to the users).

Fixes #83.

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -238,7 +238,7 @@ async function full_text_search(res, query, language, page) {
     });
     res.status(200).json({ OK: true, results: objList });
   } catch (error) {
-    log.error("Exception in GET /search/v2/", error);
+    log.error("Exception in GET /search/", error);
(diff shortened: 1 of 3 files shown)


davidascher:
let's definitely go with KISS first.

PR #115, opened 2017-05-19
36 edit apis

### LINKED ISSUE #36: Implement Edit [Case|Method|Organization] APIs
Need to get these working on top of Postgres

### LINKED ISSUE #87: Add feature set/unset API
(Can do it with the generic case/method/org save I expect, just need it for featuring front-end)

### LINKED ISSUE #99: For create/edit APIs: refresh Materialized Views for search
We need to trigger a refresh on these when the data changes. If only English text is changed:

```REFRESH MATERIALIZED VIEW search_index_en;```

Repeat as needed to cover other languages (once they are supported).

### LINKED ISSUE #108: All related content should be bi-directional
Via Jesi:

When we have related [Case|Method|Organization]s those links should be listed regardless of which way the relationship goes. Likewise, the links should be editable from either side of the relationship, so if Case-23 is related to Method-34, then viewing either of them should show the other as related content, and editing either of them should allow the relationship to be removed. Basically, regardless of where the relationship is originally defined, it should always be treated as bi-directional.

### PR DESCRIPTION
**Pull request by dethe**
Fixes #36 
Fixes #108 
Fixes #87 
Fixes #99 

Supports full JSON objects to update any of the case / method / organization trifecta, or any subset, so if you just want to set {related: true} you can send that. Will set `updated_date` automatically, so ignores that field. Also ignores attempts to set the `id` or `published_date` fields.

Has not been fully tested against the front-end because the front-end was sending old data rather than user-entered data for some fields.

### CODE REVIEW

File: migrations/migration_022.sql [hunk 1]
```diff
@@ -0,0 +1,138 @@
...
+RETURNS object_reference AS
+$$
+SELECT
+    ROW(id, 'case', title, lead_image, post_date, updated_date)::object_reference
+  FROM
+    case__localized_texts,
+    cases
+  WHERE
+    case_id = $1 AND
+    language = 'en' AND
```
davidascher: probably not right

migrations/migration_022.sql
```diff
@@ -0,0 +1,138 @@
...
+  id_1 INTEGER NOT NULL,
+  type_2 text NOT NULL,
+  id_2 INTEGER NOT NULL,
+  CONSTRAINT sorted CHECK(concat(type_1, id_1::text) < concat(type_2, id_2::text))
+);
+
+-- port starter relations from migration_013
+INSERT INTO related_nouns (type_1, id_1, type_2, id_2)
+VALUES
+  ('case', 37, 'case', 45),
```
davidascher: this belongs in test code, not in the production migration

migrations/migration_021.sql
```diff
@@ -0,0 +1,7 @@
+INSERT INTO users
```
davidascher: This doesn't belong in prod migrations, move to test code.

migrations/migration_022.sql
```diff
@@ -0,0 +1,138 @@
...
+RETURNS object_reference AS
+$$
+SELECT
+    ROW(id, 'method', title, lead_image, post_date, updated_date)::object_reference
+  FROM
+    method__localized_texts,
+    methods
+  WHERE
+    method_id = $1 AND
+    language = 'en' AND
```
davidascher: idem

migrations/migration_022.sql
```diff
@@ -0,0 +1,138 @@
...
+RETURNS object_reference AS
+$$
+SELECT
+    ROW(id, 'organization', title, lead_image, post_date, updated_date)::object_reference
+  FROM
+    organization__localized_texts,
+    organizations
+  WHERE
+    organization_id = $1 AND
+    language = 'en' AND
```
davidascher: idem

migrations/migration_022.sql
dethe: ibid

migrations/migration_021.sql
dethe: Doesn't need to be in test code either, I fixed it the right way awhile ago but forgot to delete the temporary workaround.

PR #118, opened 2017-05-23
116 richer search results

dethe:
* Added body and bookmarked to search results
* Replaced getUserIfExists with preferUser, user finding behaviour should match requireUser but will not throw an exception or create user if user doesn't exist.
* Moved code to get `bookmarked` flag into an SQL function
* More tests


davidascher:
Looks good.

PR #123, opened 2017-05-25
Upgrade hashing to rs256 -- needs synchronous change on frontend

davidascher:
Also make middleware functions less error prone -- make explicit whether they require auth or not.


test/cases.js
```diff
@@ -7,6 +7,22 @@ let should = chai.should();
 chai.use(chaiHttp);
 chai.use(chaiHelpers);
 
+// let jwt = require("jsonwebtoken");
```
dethe: Shall we delete this commented-out code?

test/cases.js
```diff
@@ -205,11 +221,16 @@ describe("Cases", () => {
   });
   describe("Get case with authentication", () => {
     it("should not fail when logged in", async () => {
-      const res = await chai
-        .getJSON("/case/100")
-        .set("Authorization", "Bearer " + tokens.user_token);
-      res.body.OK.should.equal(true);
-      res.should.have.status(200);
+      console.log("TOKEN", tokens.user_token);
```
dethe: Stray console.log?

davidascher: duh.

test/cases.js
davidascher: duh.

PR #125, opened 2017-06-01
122 refactor db

dethe:
Big changes that need to be made now or never. The main change is to use table inheritance so that cases, methods, and organizations (and eventually news and/or articles) can all be referenced and searched as `things` (the base table). This also allows us to get rid of table proliferation around the sub-tables: case__authors, method__authors, organization__authors becomes simply `authors`, and similarly for `localized_texts`.

Several functions and SQL queries were simplified (some dramatically) as a result. There are more simplifications to shake out, this round only focused on what was necessary to get the refactor running.

CAVEATS:

* Requires migration 24 to be run on the server
* Requires update to the front-end (we use `thingid` everywhere now instead of `case_id`, `thingID`, `thingId`, etc.)
* One feature I'd hoped for is not (yet) supported by Postgres: using `REFERENCES things(id)` as a foreign key constraint. That doesn't work for inherited tables.
* Uniqueness is not enforced by Postgres, so it is possible to insert a method and a case (for instance) with the same id. Some care is required, but I think our normal usage pattern is safe for this. I do have concerns about integrating updates from the old server around id collisions, but that is something we'd have to deal with in any case.


migrations/migration_023.sql
```diff
@@ -134,5 +134,4 @@ UNION
 		organization__localized_texts.language = 'en'
 ;
 
-DROP INDEX IF EXISTS idx_fts_search_en;
```
dethe: Dropping the materialized view (above) also drops the associated index, so this was just causing noise in the output with no effect.

dethe:
Very strange that the same build passes tests locally and sometimes through Travis. Makes me think maybe our tests have a race condition? Re-running the test got it to pass, but I'll need to keep an eye out for this.

davidascher:
I'm not seeing a history of failed tests above, I guess because the old failures got squashed by the rerunning. Next time, keep a log of the failures, I can look into it.

PR #127, opened 2017-06-04
add breaking test

davidascher:
@dethe, this test fails my understanding of how tags are updated.  Can you take a look?


dethe:
Good catch! I thought I'd added a test for that already, but apparently not. Should be fixed now.

PR #128, opened 2017-06-05
add currently failing test on editing issues

davidascher:
@dethe, I'm seeing some failures when editing `issue` which are interesting.  With some debugging logs added, I see:

```
POST /case/new 201 1363.581 ms - -
newThing { issue: 'new issue' }
CHANGES [ { key: '"issue"', value: 'null' } ]
PUT /case/4754 200 1265.480 ms - -
      1) Add case, then modify some fields


  1 passing (3s)
  1 failing

  1) Cases Test edit API Add case, then modify some fields:
     TypeError: Cannot read property 'should' of null
      at it.only (test/cases.js:283:25)
      at process._tickDomainCallback (internal/process/next_tick.js:135:7)
```

Note that the `value` in that CHANGES line is the _old_ value.

### DIFF
--- a/api/helpers/things.js
+++ b/api/helpers/things.js
@@ -177,7 +177,7 @@ function getEditXById(type) {
             // DELETE / INSERT any needed rows for related_nouns
             const oldList = oldThing[key];
             const newList = newThing[key];
-            newList.forEach(x => x.id = x.id || x.value); // handle client returning value vs. id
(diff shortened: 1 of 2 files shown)


davidascher:
btw the logging statement is 

```
        console.log("CHANGES", updatedThingFields);
```

in `things.js` at around line 248 (inside `if (anyChanges)`).

davidascher:
I think I found the fix.  Would love your review, @dethe.

Issue #102, opened 2017-05-02
Index all text fields for fulltext search


davidascher:
people will reasonably search for 'france' or 'montreal'. 

(probably don't need to index other bits of the geolocation IMO)

dethe:
Changed title and scope based on conversation with Amber and Jesi.

davidascher:
does that not mean create per-language tables for each field?

dethe:
We are going to have to have a longer meeting about localization soon, I think.

dethe:
At least add:

* [ ] Country
* [ ] City
* [ ] Purpose
* [ ]

dethe:
As of PR #134 we do what this issue originally asked. Need to re-raise this with Amber and Jesi for the other fields, will be complicated to do well and if we're changing metadata a bunch it will me messy to maintain.

dethe:
We increased the number of fields indexed, but indexing *all* fields will wait until we've nailed down exactly what "all fields" are.

PR #134, opened 2017-06-12
119 merge search paths

### LINKED ISSUE #107: Special-case search term "Featured"
We are moving away from faceted search. If "facet" of object type (case, method, etc.) is selected, it becomes a *filter* on the search results. The additional "facet" of *featured* will be special-cased so it is extracted from query terms and used as a filter on the `featured` boolean field. This will have implications (to be determined later) when we localize the UI.

### LINKED ISSUE #119: Merge search paths
There are currently 3 implementation paths in `/search/`, unify these.

dethe:
This overhauls our search infrastructure a bit and will require front-end changes to match. A successful search request will always return the same shape of results:

```
{
  "OK": true,
  "results": [
     // heterogeneous result set of mixed case/method/organizations 
  ]
}
```

## Arguments to /search:

* `query=<one or more words, space-separated>` (%20-separated in the URL), this is the actual search query. Default: "featured" (see below for this special case)
* `selectedCategory=[Cases|Organizaations|Methods]` this will apply a filter to only return matches if they are the right type. Default: None, return all types of results, in order of relevance.
* `resultType=map` this will cause the results to be suitable for use on the map (fewer result fields) and to ignore pagination and limits, returning all matching results. Default: any value other than 'map' (or none at all) will return regular search results.
* 'page=X` will return the Xth page of (up to) 20 results. Default: 1.

## Special case for query "featured"

If the query consists of the single word "featured" a search is not performed, but instead the current featured items are returned in the exact same format as a search result. A blank query will default to this, so the raw url `/search` doesn't actually perform a search, but returns the featured set. This was requested behaviour so that users have a way to get back to the featured list once they've navigated away.

The `/search/getAllForType` query is now deprecated and will be removed shortly.

The '/search/map' query is now deprecated and will be removed shortly.

Fixes #119 (There are still multiple paths, but they return the same shape of results and should be easier to keep in sync)
Fixes #107 
Addresses #102 by adding location.city and location.country to the search index. Adding fields not shared by all the types will take longer and be a lot more complicated.

## Known Limitation / Bug

I'm not seeing as many results come back as I expect to. I've skipped some of the tests to get back to diagnosing that later. The results look good, but there should be more matches than we're seeing. I've tried changing the weights of the different fields we index, but no luck so far. It may require me digging deeper into how Postgres full-text search actually works, but will be a job for another day.

### DIFF
--- a/api/controllers/list.js
+++ b/api/controllers/list.js
@@ -4,6 +4,7 @@ const router = express.Router(); // eslint-disable-line new-cap
 const log = require("winston");
 
 const { db, sql, as } = require("../helpers/db");
+const { supportedTypes } = require("../helpers/things");
 
 /**
  * @api {get} /list/titles Get title and id for all "things"
@@ -43,7 +44,7 @@ router.get("/titles", async (req, res) => {
     // console.log("organizations: %s", retVal.organizations.length);
     res.status(200).json({ OK: true, data: retVal });
   } catch (error) {
-    console.trace("Exception in POST /case/new => %s", error);
+    console.trace("Exception in POST /list/titles => %s", error);
     return res.status(500).json({ OK: false, error: error });
   }
 });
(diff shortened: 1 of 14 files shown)


davidascher:
"A blank query will default to this, so the raw url /search doesn't actually perform a search, but returns the featured set. This was requested behaviour so that users have a way to get back to the featured list once they've navigated away." -- that feels like a user request which can be handled by the front-end, not complicate the API.

PR #140, opened 2017-06-14
139 clear cache

### LINKED ISSUE #139: Clear cache on PUT (or POST?)
If we flip the featured bit, it looks like nothing happened because the search is cached. Same for adding more tags, presumably.

dethe:
Fixes #139 

The last test is the important one, the rest of the test changes are just bringing search tests in line with the code style for case / method / organization tests using async / await.

Clears the cache when we add a new "thing".


api/controllers/case.js
```diff
@@ -164,6 +164,8 @@ router.post("/new", async function postNewCase(req, res) {
   }
   */
   try {
+    const cache = require("apicache");
```
davidascher: I don't think this does anything w/o an actual call to `cache.clear()`

dethe:
Aside from accidentally checking in swagger noise.

PR #146, opened 2017-06-15
135 search results

### LINKED ISSUE #84: Search api needs to return # of hits-not-returned
We need to be able to say "showing 30 hits out of  133" and compute the # of pages.  So we need "total_hits" added to the return payload on `/search` (and ideally all paginated endpoints)

### LINKED ISSUE #135: Figure out why we're not seeing as many search results as expected
Start with the `.skipped` tests in search.js.

### LINKED ISSUE #141: cases created without a topic or issue will return a string null instead of a null for those fields
create a case without specifying topic or issue.
the case returns with `"null"` as a value for those.

note: the point is it's a string, not an actual `null` in JSON.

dethe:
Now indexing all objects for search, instead of only about 1/4 of them. Also:

* Returns `pages` and `total` in search results (but not map-view results) showing how many matches there are overall in raw numbers and pagination respectively.
* Fixes "null" -> null both for historic data and going forward

Fixes #84 
Fixes #135 
Fixes #141 

Requires migration 28

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -57,7 +57,7 @@ const singularLowerCase = name => name.slice(0, -1).toLowerCase();
 // like it says on the tin
 const filterFromReq = req => {
   const cat = singularLowerCase(req.query.selectedCategory || "Alls");
-  return cat === "all" ? "" : `AND things.type = '${cat}'`;
+  return cat === "all" ? "" : `AND type = '${cat}'`;
 };
 
 const queryFileFromReq = req => {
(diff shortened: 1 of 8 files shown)


api/controllers/search.js
```diff
@@ -121,7 +121,12 @@ router.get("/", async function(req, res) {
       offset: offsetFromReq(req),
       userId: req.user ? req.user.user_id : null
     });
-    res.status(200).json({ OK: true, results: objList });
+    const total = Number(objList.length ? objList[0].total || 0 : 0);
```
davidascher: wow that is obtuse.  I can't simplify it though.

PR #149, opened 2017-06-16
148 hidden things

### LINKED ISSUE #148: Add "hidden" flag to objects
As a "delete-lite" option

dethe:
Support for marking things "hidden" to remove them from search results and lists. You can still pull them up by id for view or edit if needed.

Fixes #148 

Also cleaned out some files that were no longer needed and other misc. cleanup.

Requires migration 29


api/sql/thing_by_id.sql
```diff
@@ -15,6 +15,8 @@ WITH authors_list AS (
 full_thing AS (
   SELECT
     ${table:name}.*,
+    COALESCE(location, '("","","","","","","","","")'::geolocation) AS location,
```
davidascher: this seems unrelated to hiding - what is it for?

api/sql/thing_by_id.sql
```diff
@@ -15,6 +15,8 @@ WITH authors_list AS (
...
+    COALESCE(lead_image, '("","",0)'::attachment) AS lead_image,
```
davidascher: Same here

migrations/migration_029.sql
```diff
@@ -0,0 +1,56 @@
+-- Fix instances of "null" vs. null
```
davidascher: this is duplicating a bunch of `_028`, which doesn't seem right.

api/sql/thing_by_id.sql
dethe: Ensures that we always return an empty location rather than null. We do this everywhere else, but I'd missed this one, added for consistency.

api/sql/thing_by_id.sql
dethe: OK, had missed more than one.

migrations/migration_029.sql
dethe: Changes in _28 hadn't caught everything I thought they would, so this fixes the fix.

PR #232, opened 2017-10-28
return organization name

### LINKED ISSUE #231: Return organization name for user profile

dethe:
Fixes #231 

Instead of returning the id of an organization, this will return a structure of `{name: , id: }`. If there is no associated id the structure will still be there, but will be `{name: "", id: null}`.

@andreadelrio Will this work for you. You will need the name for display, but the id to build a link.


dethe:
We no longer need orgs for users

PR #237, opened 2017-11-14
236 map category

### LINKED ISSUE #226: Map: show filtered-out content as disabled pins
Just like search results, keep all content in the map, but disable types that are filtered out (i.e., if we are filtering only cases, show methods and orgs in grey and not clickable).

### LINKED ISSUE #236: search with resultType=map and selectedCategory doesn't return searchmatched true
From what I can tell, the query:

https://api.participedia.xyz/search?query=france&resultType=map&selectedCategory=Case

should return some entries with `searchmatched` set to `true`, but doesn't (they're all `false`).

This blocks https://github.com/participedia/frontend/issues/478

dethe:
Reworked how the search API works with map results to make it more consistent and reliable, and added some tests to catch regressions.

The API allows you to over-ride the page size (`limit=[X]`) for non-map queries (map queries always return everything). Also added `searchhits` to results in map search to know at a glance how many of the results should be clickable.

Fixes #226 
Fixes #236

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -82,7 +82,6 @@ const queryFileFromReq = req => {
   } else if (resultType == "map") {
     queryfile = SEARCH_MAP;
   }
-  console.log("query file: %s", queryfile.file.split("/").pop());
   return queryfile;
 };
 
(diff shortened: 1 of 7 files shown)


api/controllers/search.js
```diff
@@ -86,7 +87,17 @@ const queryFileFromReq = req => {
 
 const offsetFromReq = req => {
   const page = as.number(req.query.page || 1);
-  return (page - 1) * RESPONSE_LIMIT;
+  const limit = limitFromReq(req);
```
davidascher: this line seems useless?

dethe: Oh yes, lost my place mid-refactor. Good catch.

PR #289, opened 2018-07-15
use admin rights from our db instead of auth0 when saving data

dethe:
Server-side part of fix for frontend issue 800.


dethe:
Users with the `isadmin` flag in db can update `hidden` and `featured` but the fake user for the tests cannot. Fuck it, pulling this into master anyway.

PR #294, opened 2018-08-19
291 new templates

### LINKED ISSUE #291: New APIs for backend migration
Each page (or most of them, anyway) should support access as html, html fragment, JSON, XML, or CSV. All text in each page template must be localized values.

#### Arguments

* `?query=[search terms]` for full text search
* `?page=[pageno]` for pagination of search results
* `?display=[list | grid]` layout of search results
* `?type=[all | cases | methods | tools | organizations | resources | news | users]` filter, only returns selected type
* `returnType=[null | map]` return search results for display or for map
* `resultType=[json | csv | xml | html | htmlfrag]` to get back different formats from the API

#### Search path

* [ ] `/` Homepage and search results. 

#### Static pages

Static pages include `about`, `teaching`, `experiments`, `research`, `legal`, and possibly `admin`.

* [ ] `/[page]`
* [ ] `/[page]/edit`
* [ ] `/[page]/localize`
* [ ] `/[page]/edit-localize`

### Article pages

Articles are all the database thing types: cases, methods, teaching tools, users, organizations, news, etc.

* [ ] `/[thingtype]/[id]` View article
* [ ] `/[thingtype]/[id]/edit` Edit article
* [ ] `/[thingtype]/[id]/localize?[from=en;]to=es` Localize article
* [ ] `/[thingtype]/localize?[from=en;]to=es` Localize edit template
* [ ] `/profile` redirects to `/users/[id]` for the current user's id if they are logged in

## deprecated paths
* [ ] `/cases` deprecated search filter, replace with `/?type=cases`
* [ ] `/methods` deprecated search filter, replace with `?type=methods`
* [ ] `/organizations` deprecated search filter, replace with `?type=organizations'

Implementation note: At the moment, I can't see a difference between the single-column localize interface and the edit interface, so there may not be a need for edit-localize paths.

dethe:
Implement templates for backend using Handlebars (mustache-based) for templating. Must support layout, page templates, and partials.

Fixes #291


dethe:
There is still a lot to be done on the [Big List of Tasks](https://github.com/participedia/api/issues/291#issuecomment-422909869), especially saving cases and getting tests passing again, but I'm going to merge this now because it is the basis for work going forward, and it does in fact implement the original request, it's just that this issue became a catch-all for "things that need doing". Remaining tasks on the Big List can be factored off into separate issues, if they don't already have issues.

PR #333, opened 2018-11-30
footer markup and styling

ascott:
Implements responsive footer, updates the font paths in style.css and adds normalize.css to make all browsers render elements consistently.

This gif shows the scaling:
participedia-footer-responsive-new

for review @dethe


ascott:
@jesicarson @dethe i updated the styling to reflect jesi's mobile footer design, and added a new gif. let me know what you think!

jesicarson:
Snazzy 👏

public/css/vendor/normalize.css
```diff
@@ -0,0 +1,349 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
```
dethe: My feelings about normalizers are that, by and large, we should incorporate them in our code. That is, instead of resetting to a sane default and then over-riding for our site's needs, we should set to out site's needs right off the bat.

That said, I don't see a lot of overlap between what is in this reset and what we define ourselves, but at any point that we find ourselves overriding the reset, we should fork it and make the reset the basis of our styling. A good example of that is here: https://igoradamenko.github.io/awsm.css/

No change needed, and if you disagree with the above, please let me know, I don't want to dictate how things are done.

views/partials/footer.html
```diff
@@ -1,36 +1,75 @@
-<div class="footer-component">
-  <div class="mobile-footer pb-4 d-md-none d-lg-none d-xl-none container">
-    <div class="pt-4 row">
-      <div class="col-5 offset-1"><a href="/about"><span>About</span></a><a href="/research"><span>Research</span></a><a href="/teaching"><span>Teaching</span></a></div>
-      <div class="col-6"><a href="/news"><span>News</span></a><span class="open-help">Help &amp; Contact</span><a href="/legal"><span>Terms of Use</span></a></div>
+<footer>
```
dethe: This looks awesome.

dethe:
Merge at will :shipit:

PR #339, opened 2018-12-04
responsive header and sidebar menu

ascott:
Implements header and sidebar/hamburger menu.

- sidebar menu is opened/closed via css. i added a close link to the menu, since it's not possible (afaik) to make the background clickable to dismiss via css only. @jesi/@dethe let me know if you want to handle this differently. i noticed a ux issue with this solution: it uses a link hash (`href="#sidebar-menu`) to trigger the opening of the menu. if a user navigates from the menu to a new page and then hits back, they will be taken to the url with the hash #sidebar-menu, which will render the page with the menu open. i think ideally we would always want the page to load with the menu hidden.

todo: 
- [ ] implement login link/handling

participedia-responsive-header


jesicarson:
Hi @ascott, can the hamburger button itself become the "close" button when it is open (a la [google play](https://play.google.com/music/listen?u=0#/home))? Also, according to this design, the hamburger opens below the top nav bar, but this makes me think I need to update that design https://github.com/participedia/frontend/issues/853#issuecomment-426301233 - thoughts?

ascott:
hey @jesicarson, i was looking at the .xyz site and was copying the functionality/style for the sidebar menu from there, where it stretches the full height of the window. i can update to match the design so that it is displayed below the top nav bar, and uses the hamburger icon to close the menu as well.

ascott:
@jesicarson i updated the gif with the changes to the sidebar menu. let me know if you would like to see other changes.

jesicarson:
@ascott Only other changes would be to add the quick submit and profile/login buttons to the hamburger menu per the design guidelines. I am also considering placing a light grey circle around the hamburger icon when it is open to indicate that you click it to close... However this is good for now, thank you!

Question though, with this new system when you click outside the hamburger menu when it is open, it won't close? Or is the close button an extra measure? Because currently on xyz it closes when you click anywhere outside the menu. Sorry if you've explained this already! Just want to double check. Ideally you click outside the menu to close it I think. 

And, lastly, there appear to be some grey rectangle shapes appearing around the logo/search bar area when you're scaling down in size - is that just a gif glitch or is something odd going on there? Thanks!!

ascott:
@jesicarson i will add the quick submit and profile/login buttons. 

re: clicking anywhere to close the menu: this isn't possible with CSS only, but i can add this functionality with JS. @dethe's approach is to do as much as possible with HTML/CSS only, and add JS to enhance the UI.

re: the grey rectangle shapes: yes, these are just gif artifacts. please disregard!

jesicarson:
@ascott thanks for the clarification!
I think the design as is is sufficient, and let's maintain the dark grey (Grey C #808080) on the hamburger menu icon when it is open, in addition to its hover state when closed. Thanks!

ascott:
@jesicarson just to clarify, you want the hamburger menu icon colour to be `#808080` when the menu is open, as well as on hover? what should the hover be when the menu is open? back to `#fff`?

jesicarson:
Oh didn’t think of that. Back to white works!

PR #345, opened 2018-12-14
Case Edit Form - HTML fields

ascott:
- adds buttons/logic for quick submit vs full submit

- moves handlebars helpers to it's own file to keep things organized, adds a bunch of helpers

- adds html for fields
-- boolean/checkbox
-- text
-- date
-- number
-- single select
-- textarea
-- edit-list-ranked
-- edit-media-link-list
-- location (adds single text field for all location attributes, need to add js for autocomplete)

- adds submit and floating submit buttons

- adds placeholders for JS driven fields

- adds temporary handling for select options until the api changes for this are merged in

participedia-case-edit-2


dethe:
Ah, I see what you mean by a `full key` now. Is it possible to render the form either way, but to hide/show using CSS to switch between full and quick views?

ascott:
@dethe we are using css to show/hide the correct fields depending on if it's full submit or not. we are using the key on the article to set a data attribute called `submit-type` in order to set the correct css styles. https://github.com/participedia/api/pull/345/files#diff-8df5c69eb793c611517fc5a299761c7dR3
we could add the key to the parent node rather than the article node if that makes more sense...

ascott:
@dethe moved the param check for full vs quick to a handlebars helper. think it makes more sense here. https://github.com/participedia/api/pull/345/files#diff-5cd984740a5faccec00678b105dc5be0R118

PR #350, opened 2018-12-19
multi-select and ranked multi-select with warning modal

ascott:
- adds multi-select component that supports ranking/re-ordering and max number of items
- adds polyfill to support drag and drop on touch screen devices
- adds dialog/modal for user feedback

edit-multi-select-2

screenrecording_12-18-2018 22-57-32 2018-12-18 23_19_02


ascott:
@jesicarson here are screenshots of the ui for this field, let me know if you'd like to see changes. 

screenshot 2018-12-19 10 57 13

screenshot 2018-12-19 10 56 46

screenshot 2018-12-19 10 59 10

ascott:
thanks @jesicarson! will update with these changes.

ascott:
hey jesi, thanks for the feedback. here are updated screenshots:

screenshot 2018-12-19 19 19 30

screenshot 2018-12-19 19 19 05


In terms of the modal, both the X icon and close button are widely used. The X is more subtle than the red button. Alternatively, we could change the colour of the close button so it's not so distracting. Here's a screen shot with the X in the top right corner. 

screenshot 2018-12-19 18 54 30

jesicarson:
@ascott beauty!! this is so satisfying you have no idea lol

Ok, the forms look perfect.

The modal box... I think the X works for this instance, and for something like the file types you're adding aren't supported. More of an FYI than an urgent warning. Called "info popup" in mockup below.

I think we **should** use the "Warning" title text and red cancel button **only** in the instance when someone may navigate away from un-published changes. 

The only changes I would make to the modal box would be to: 
- [ ] standardize the width and placement on the screen for desktop and mobile. I am open to your suggestion for mobile, but have designed for desktop to have modal boxes match the width of the form fields/center column of text on the site. In terms of placement, centered on the screen makes sense to me? On mobile, the popups could fill the entire screen with an x in the corner, rather than overlay the center of the screen, OR use the same overlay system with black background as desktop with width matching width of the content, I am open to whichever is faster/easier to do.  
- [ ] standardize where the text box begins and ends inside the white bounding box (20px from the top right corner and left edge) and (20px to the left of the X), as per the mockup below found in the vector UI kit doc. 
hamburger-feature tour-publish-info popup

jesicarson:
Also, I like the opacity of the black modal shown in your last screenshot - if its different than 60%, let me know, as I prefer it and will update the design doc. Thanks!

jesicarson:
oh, and the X should always be dark grey.

ascott:
@jesicarson 🎉 

- the background overlay opacity is 75%
- changed X icon color to #808080 (dark grey)
- i made the modal width match the content width for desktop and mobile screens
- i haven't implemented the publish warning yet, but will refer to these mocks when i do.

screenshot 2018-12-21 12 33 41

screenshot 2018-12-21 12 31 55

dethe:
Merge at will :shipit:

PR #354, opened 2018-12-27
Case Edit - rich text editor

ascott:
- uses [browserify](http://browserify.org/)/[watchify](https://github.com/browserify/watchify)/[esmify](https://github.com/mattdesl/esmify) to bundle js files so we can use NPM modules on the frontend
- uses [QuillJS](https://quilljs.com/) for rich text editor

screenshot 2018-12-27 14 29 25

(note: comparing against `alanna-case-edit-form-multi-select-ranking` until it is merged into master)

@dethe @jesicarson


dethe:
Oh, I see, there are a ton of changes in package-lock.json that aren't in package.json.

dethe:
That’s all great, no urgency for the remaining bits, I just want to make sure I know where we’re at on the various parts of the larger issue, and that we’ve documented it for th rest of the team.

All good, merge at will. :shippit:

jesicarson:
Yay for closing issues! Nice work team. Agreed no rush on remaining elements. 
My only question/comment is about clicking links when editing: for the option to visit the link, either it must open in another tab, or present the warning about navigating away from unpublished work. Thanks!

ascott:
@jesicarson the links from the editor will open in new tabs.

PR #356, opened 2018-12-29
Case Edit: Location autocomplete, edit-link-list, edit-media-list interactions

ascott:
- implements autocomplete location fields
---- open question: can i send up lat/lng in decimal format rather than degrees? google maps returns decimal lat/lng so would be easier to return that format rather than convert on the frontend.
participedia-location-autocomplete

- implements handling for adding new field for link list
participedia-link-list

- implements handling for adding new media link list sets
participedia-media-link-list

@dethe @jesicarson


ascott:
closing, opened with wrong branch.

PR #357, opened 2018-12-29
Case Edit: Location autocomplete, edit-link-list, edit-media-list interactions

ascott:
- implements autocomplete location fields
---- open question: can i send up lat/lng in decimal format rather than degrees? google maps returns decimal lat/lng so would be easier to return that format rather than convert on the frontend.
participedia-location-autocomplete

- implements handling for adding new link sets (including title and attribution fields)
participedia-media-link-list

@dethe @jesicarson


dethe:
I'm not sure what we were using for choosing location information in React. We have been storing text strings of degrees and minutes, but there is a Geometry(POINT) type in Postgres for storing (longitude, latitude) as floating point. I can work out converting our existing data. We'll need to make sure the map widget can take floats, since that's really the only place we use the latitude and longitude.

In other words, I'm happy to store whatever works for the front-end.

jesicarson:
@ascott 
Great thanks! Except the visual is an X, inside the grey outline of the added value box, not a garbage icon. This mockup shows the consistent use of the X for both selected values and added files/media (that said, if it's a pain to change this, not the end of the world and we can return to it later):
form - added values and media

Regarding the evaluation report/links - Confirmed, these should indeed have attribution fields. Scott has updated the [case fields editor spreadsheet](https://docs.google.com/spreadsheets/d/1zAkNVa-mjMsMqtIrjYL-j96fP4LTGOTFjs668WwxxRI/edit#gid=1762179507) to reflect this (row 62 onward, highlighted red). Apologies for not catching that before you built the link list!

jesicarson:
@ascott on second thought, I like how clean it is with the icon outside the box - it avoids resizing the inner fields as well. I'd be happy to keep as is, and just replace the garbage icon with an X. Although, I'm curious about how it works on mobile, if the fields are close to the edge of the screen, does it resize them to make space for the icon?

PR #364, opened 2019-01-06
Case Edit -- unpublished changes dialog and other smallish changes

ascott:
- adds browser dialog to warn user when navigating away from/reloading the page before submitting changes
---- @jesicarson a note on the design: The way that browsers handle the event when a user is about to leave the page doesn't allow us to show a custom modal but it does have native support for this functionality and shows a system dialog in both cases of leaving the page and reloading the page. The dialog is only shown if the form data has been changed. 

screenshot 2019-01-04 18 33 57

screenshot 2019-01-04 18 33 21

-----
other changes
- don't show the map on edit form page
- generalize case edit css for use on all edit form pages
- don't init location, multi-select, or rich text editor if those fields are not on the page (to make generalizable for other form pages)
- make sure long urls in rich text editor don't break layout on mobile devices
- limit character input on textarea to 280 characters
- fix issue with long labels/p tags breaking mobile layout


jesicarson:
@ascott native modal is fine. Thanks!

PR #367, opened 2019-01-09
291 more static

dethe:
Still not sure why these two fields did not get added earlier, I ran the same function on both of them as I did for the other fields. The individual parts of the function are working: I can extract the core select and get the expected values back, but the function as a whole doesn't work (just for these two fields).

So I extracted the values into long-form INSERT statements like an animal. Seems to work.

Onward to saving modified cases!


ascott:
👍

dethe:
Forgot to note, this requires `npm run migratelocal 28`

PR #407, opened 2019-01-10
Case Edit -- Edit Submission Details 

ascott:
- implements editing of submission details using an autocomplete drop down for author names

participedia-edit-submission-details
@jesicarson, i simplified the edit button until we implement the other admin functionality. does that work for now? 

@dethe, I need a couple things from the backend to complete this
- [x] add isAdmin flag to json blob
- [x] add author data to json blob (an array of objects with user_id and name keys)


jesicarson:
@ascott works for me!

dethe:
@ascott I have authors as a list of who made changes when, but you just want id/name pairs with no repeats, yes?

dethe:
If you use the `?user=[id]` hack, the user object does have `isadmin` as a boolean.

dethe:
PR #410 adds the `authors` field. It will be populated with names and ids of all the users if the current user is an admin, or an empty list otherwise.

PR #408, opened 2019-01-10
Help & Contact Widget

ascott:
- implements the help & contact widget
participedia-help-contact

the gif makes the box shadow around the widget look like a solid gray, but it is actually a more subtle gradient/shadow

screenshot 2019-01-10 11 18 34

@dethe i've hardcoded the faq content for now, but we can swap it out in a subsequent PR with the translated copy when that gets added to the json blob.

@jesicarson @dethe


jesicarson:
@ascott looks great!

PR #410, opened 2019-01-13
add authors list to case edit when the user is an admin

dethe:
Adds list of {id, name} for all system users to case data as `authors` if the current user is an admin. If the current user is not an admin, returns an empty list as `authors`.

Requires `npm run migratelocal 30`

Requested as part of #407


ascott:
thanks! i will incorporate this into my [branch](https://github.com/participedia/api/pull/407).

PR #415, opened 2019-01-17
Reader View

ascott:
- adds/updates templates for all reader views
- adds case reader view
- adds meta/social tags for article pages (for search/twitter/fb cards)

[mobile screenshot](https://user-images.githubusercontent.com/130878/51303086-22986d80-19e9-11e9-98ab-bc7c0c1cfe9c.png)
[desktop screenshot](https://user-images.githubusercontent.com/130878/51303084-21ffd700-19e9-11e9-8963-bc48de4991c8.png)

participedia-reader-view

@jesicarson @dethe


jesicarson:
@ascott looks great! Couple small visual things:
- The titles on the images should only appear/slide up on mouseover on desktop or on touch on mobile. And there should be a full screen icon (which can be implemented later). 
- The drop down arrow on case data looks too large, can we try the small arrow (use the same size as the left and right arrows beside the photo thumbnails maybe) and a bit more space between the horizontal line above the text? Also, add another horizontal line below the case data text and drop down arrow, so its more of a box. See screenshot, but don't worry about the submission details, that can stay out of a drop down as you have it. Thanks!
screen shot 2019-01-21 at 9 28 04 am

jesicarson:
@ascott It also looks to me that the type sizes on the reader view need to be adjusted - those main titles (H2) should be larger than the submission details / case data (H4). 

Maybe you could also post a static screenshot, as it's a bit hard to see with all the scrolling in the gif. Or let me know when staging is ready, as that would be the best way to review. Thanks!

ascott:
@jesicarson about this comment 
>  those main titles (H2) should be larger than the submission details / case data (H4).

are the main titles you are referring to the headings within the body/narrative text? if so those headings are using h3 tags rather than h2 tags. the body/narrative text is stored as html and so these headings/html are controlled by the user. we could alter the size of h3 tags only for the body/narrative text via css, but it doesn't seem ideal to overwrite expected behaviour for a tag like that. since the body/narrative field allows a user to use html to format their text, we have less control over how to format that area since the html could be in any format/structure.

jesicarson:
@ascott yes the headings within the body. for the new version of the site, can we change the default for the narrative headings to be h2? per the latest design guidelines https://github.com/participedia/frontend/issues/853#issuecomment-400787153
screenshot: 
screen shot 2019-01-22 at 11 20 38 am

jesicarson:
^ Note, the blurb above "Problems and Purpose" is the text from the short description field.

ascott:
thanks @jesicarson. i created a PR with changes here: https://github.com/participedia/api/pull/420

PR #420, opened 2019-01-22
Reader view design updates

ascott:
based on @jesicarson's feedback here: https://github.com/participedia/api/pull/415

- on slideshow, only show title text when hovering over slideshow container
- adjust styling on accordion header
- add floating edit button
- style h3's within body content to look like h2 headings

participedia-reader-view-slideshow

screenshot 2019-01-22 11 38 53


jesicarson:
@ascott thanks! 

On the case data drop down: Could you add a bit more space between the text and the horizontal lines, just looks a little squished, an additional 5 px above and below the text should do it. 

And when you say "style h3's within body content to look like h2 headings" - can you change whats in the form/ user entered narrative box headings to actually use h2, so you don't have to overwrite the style from h3 to "look like" h2?

ascott:
@jesicarson ok, i'll make that change to the case data dropdown styling.

for the h3's in the body/narrative field: we would have to write a script to alter all of the existing article data in the database to change the h3's to h2's. i would suggest that we may want to wait until post launch to tackle this. @dethe do you have any thoughts on this?

jesicarson:
I see, yes the heading issue can wait until after launch.

ascott:
updated case data/accordion dropdown header:
screenshot 2019-01-24 18 01 06

PR #421, opened 2019-01-24
Home/Search Page

ascott:
hey @jesicarson & @dethe, here is the home/search page.

todo: 
-  [ ] update results for each type (cases, methods, organizations) -- we currently only have results with all types in the json, need to add the filtered types as well
- [ ] test sort by functionality (took a quick look and it doesn't look like this is implemented in the backend yet, @dethe but plz let me know if i've missed it somewhere.)

desktop
home-desktop

mobile
home-mobile

responsiveness - scales from 4 cards across, to 3, to 2, and to a single column on the smallest screens
participedia-home-desktop-to mobile

desktop
participedia-home-desktop

mobile
participedia-home-mobile


jesicarson:
@ascott 
First of all... YAY YAY YAY looks so good!
couple tweaks:

- Can you please add an additional approx 3-5px of leading between the lines of text on the tag line, so the spacing between lines of text appears similar visually to the leading on the entry titles on the tiles below?

- Is it possible to remove the dotted lines on the map? they really emphasize that it is duplicating when it is zoomed out... another option would be to default to a more zoomed in view so the lines arent so prominent. thoughts?

- The "sort by" drop down arrow should match the pagination arrows, but be on its side, per the Vector UI kit in the guidelines. This is the same arrow used in drop down menus in the forms, and on smaller accordion drop downs like the data drop down on reader pages/ nested sub-lists on about page - for visual consistency throughout. 

- Are you showing the smallest screen as in that's what it would do on mobile? Or does the map not load on mobile? I kind of like having the map available on mobile, but we'd need to do something about how it works (ie: there's no command button). Thoughts?

THANKS!!

ascott:
thanks @jesicarson :) i'll make these tweaks.

> Is it possible to remove the dotted lines on the map? they really emphasize that it is duplicating when it is zoomed out... another option would be to default to a more zoomed in view so the lines arent so prominent. thoughts?

yes, i can remove the dotted lines

> Are you showing the smallest screen as in that's what it would do on mobile? Or does the map not load on mobile? I kind of like having the map available on mobile, but we'd need to do something about how it works (ie: there's no command button). Thoughts?

yes, the smallest screen is the smallest likely mobile screen (320px wide). currently the map is still shown on mobile and it works the same way as on the desktop. would you prefer that it is not shown on mobile? i'm not sure what you mean by  "no command button"...

jesicarson:
@ascott 

Great thanks!

I thought I saw an overlay on the gif that said "use command + scroll to zoom" or did i imagine that? lol

I'm happy with the map on mobile, as long as it doesn't eat a lot of data - can we mitigate this? Touch to activate perhaps?

ascott:
> I thought I saw an overlay on the gif that said "use command + scroll to zoom" or did i imagine that? lol

oohhh, i know what you mean now. that was because i was using my track pad to scroll, so on mobile, this wouldn't be an issue.

in terms of data, we are using the same google library for the map that we are using for the location autocomplete on the edit forms, so we have to load it anyway for that use case. we should def do a speed/performance audit post launch and decide where we'd like to/need to make changes, but at the moment, it is not a huge data hog.

jesicarson:
Sounds good!

dethe:
Filtering by type should work in the backend, I will check on that.

PR #426, opened 2019-01-28
396 login

dethe:
Not perfect, but basic login seems to be working. There's a popup that reports renewing the token failed which I still need to address, and logout  needs some attention, but if you (@ascott ) can try it and let me know if it works for you too, then at least I know I'm on the right track. Let me know of any issues you have.


api/helpers/handlebars-helpers.js
```diff
@@ -68,6 +68,9 @@ module.exports = {
     }
   },
   hasValue: (article, name) => {
```
ascott: this function should return a boolean. i'm only using it [here](https://github.com/participedia/api/blob/master/views/partials/edit-richtext.html#L3) currently, but wanted to make it reusable in case it was needed for another field. if article is null for some reason, i think it would be preferable to return false rather than a string.

public/js/authservice.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+  var webAuth = new auth0.WebAuth({
+    domain: 'participedia.auth0.com',
+    clientID: 'lORPmEONgX2K71SX7fk35X5PNZOCaSfU',
+    responseType: 'token id_token',
+    scope: 'openid profile email picture user_metadata',
+    redirectUri: location.origin + "/redirect"
+  });
+
+  var loginBtns = document.querySelectorAll('.loginButton');
```
ascott: i've been using the pattern of using class names prepended with `js-` to distinguish classes used for style and classes used for behaviour, so i would suggest changing this to to `js-login-button` for consistency.

ascott:
thanks for this dethe! i can confirm i'm also seeing the timeout pop up and logout doesn't seem to be working, but login does seem to be working.

ascott:
i'm going to merge this in so i can get started on some user related work on the frontend.

PR #430, opened 2019-02-02
User profile view and edit pages

ascott:
- adds user profile and user profile edit pages
- refactor some form inputs to be shared between article edit pages and user/profile edit
- refactor tabs and cards to be reused on user profile view and home/search view
- updates header and sidebar with profile img/button instead of login links when user is logged in

user-view template
screenshot 2019-02-01 18 01 46

user-edit template
user-edit

@dethe @jesicarson


jesicarson:
@ascott this looks perfect! tiny tiny detail, can you make the list view icon the same height as the grid view icon?

Also, questions for you and @dethe - what elements of the user profile (if any) should be localizable? are people going to translate each others bios? That seems unlikely. At any rate, this is sort of a reminder that the "Admin Options" button may need to be included somewhere, as I suppose Admins will have to be able to edit things? Not sure. Or will can admins simply visit a users profile and see the edit profile button? Thanks for helping me think this through. Maybe I'm overcomplicating in my mind :)

ascott:
thanks for the review @jesicarson, i'll fix the icon sizing 👌 

my thinking in terms of admin options: we have added admin only ui to update contributor/author info on cases/methods/orgs edit forms currently, but other admin/translation tools will have to come in our p2 launch. before then, dethe and i will be able to make needed admin only changes without the ui, and we can prioritize what admin ui/tools are most important after the p1 launch.

jesicarson:
Sounds great, thanks @ascott!

dethe:
Sorry for the late reply.

> Also, questions for you and @dethe - what elements of the user profile (if any) should be localizable? are people going to translate each others bios? That seems unlikely. At any rate, this is sort of a reminder that the "Admin Options" button may need to be included somewhere, as I suppose Admins will have to be able to edit things? Not sure. Or will can admins simply visit a users profile and see the edit profile button? Thanks for helping me think this through. Maybe I'm overcomplicating in my mind :)

I don't think bios should be localizable. Everyone can write their bio in their native language. But it's not really my call, if the executive committee want bios to be localizable and think there are people to do the translating, we'll make it so.

Issue #369, opened 2019-01-09
Case model saving/validation


ascott:
Validate submitted Case data and save changes:

* [x] Accept `JSON` 
* ~~Accept `application/x-www-form-urlencoded` POST data (or `multipart/form-data`?) for both JS and HTML-only scenarios (defer to post-launch)~~
* [x] Missing values will be ignored so the front-end can send partial values, only what has been updated.
* [x] Extra values (keys we don't support) are discarded without error.
* [x] Some values can only be changed by an admin, changes by non-admin will also be discarded.
* [ ] SQL commands, HTML/CSS/JS in plain string will be stripped silently.
* [ ] Rich Text (HTML) will go through a filter that only allows whitelisted elements and attributes. All JavaScript, CSS, and SQL will be stripped silently.
* [x] Posting changes to an article ID which does not exist, is hidden, or is the wrong type (e.g., ID is for Method, but matches a Case) results in a `404 error`.
* [x] Creating a new article without minimum content (has to have at least a title) results in a `400 error`.
* [x] Sending changes without being logged in results in a `403 error`.
* [ ] Errors will have the specified error code and be in the requested format (json, html).
* [ ] Some values can be submitted in multiple formats. A list of linked methods can be sent as an array of {id, name} pairs, or as an array of ids and either form should be accepted.

Further issues will add new requirements. When we implement the versioning API we will pass a hidden version value to return with the POST data, and if the version does not match the database version of the article it will result in a `409 error` because the article was changed between reading and submitting and those changes would be overwritten. This is out of scope for this particular issue though.

PR #454, opened 2019-03-03
369 save case

dethe:
Addresses most of #369, but still needs some work. 

Requires `npm run migratelocal 32` or `npm run resetlocaldb`


ascott:
Hi Dethe, when I send a POST request to `/case/{id}` with the quick submit data ([gist](https://gist.github.com/ascott/a5fef0078946b3da5da94edd63817f85)) I get this error ([gist](https://gist.github.com/ascott/b86712c7ab945cf90a2dcea2fb7db08e)). Any ideas? Should I be sending up an `original_langauge` key with the POST data?

dethe:
Hi Alana,

Sorry, thought I tested that. Yes, it does expect an `original_language` field. I could default it to "en" when unspecified, but once we get going with localization that wouldn't be a good idea.

ascott:
made a PR to get things working for the frontend: https://github.com/participedia/api/pull/457

PR #457, opened 2019-03-05
Case saving -- fix errors and make POST /case/id return json

ascott:
hey @dethe, i know you are on vacation, so feel free to ignore, but just wanted to make a PR for these changes to your branch before i merge into master. i made the following changes to get things working so i can build the success and error states on the frontend.
- i added hidden fields on the front end that the api is expecting but that have not been implemented yet
- commented out a db transaction that was failing despite sending the correct data from the client, i'm hoping this is something you can look into when you have a chance
- make POST /case/id return json so the client can handle success and error states asynchronously rather than waiting for a full round trip from the server.

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -155,7 +155,12 @@ function getUpdatedCase(user, params, newCase, oldCase) {
     updatedCase.hidden = as.boolean(newCase.hidden);
     updatedCase.original_language = as.text(newCase.original_langauge);
     updatedCase.post_date = as.date(newCase.post_date);
-  } else {
+  } else if (oldCase) {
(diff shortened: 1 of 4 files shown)


dethe:
Hi Alanna, sorry for the delay, we've been alternating between running errands and being KO'd by jet lag. Glad you moved ahead with this.

Which db operation was failing? I'll take a look when I get a chance to sit down with my Mac.

ascott:
hey, no problem. hope you are having a good trip. i left a comment in the code here so you can see what was happening https://github.com/participedia/api/pull/457/files#diff-07780e70b8cc879c37de679a75e0e6bbR350

dethe:
Great, thanks!

PR #458, opened 2019-03-06
save article edit form error and success handling

ascott:
- update json from server to return array of error validation strings rather than an object
- use a11y-dialog-component from npm rather than vendor file now that we have browserify bundles implemented
- on save success, redirect to reader view
- on save error, show modal with list of validation errors
screenshot 2019-03-05 16 43 21
@jesicarson please let me know if you have changes to this UI


jesicarson:
@ascott looks great thanks!

PR #460, opened 2019-03-06
replace Fakt with open source font Inter

ascott:
Inter -- https://rsms.me/inter/

- FontFaceObserver doesn't appear to support loading ttf files, so waiting for fonts to be loaded before rendering wasn't working properly
- replace Fakt with open source font Inter
- update a few font properties to make it look as close to Fakt as possible
- we only need 2 weights for Inter and the files sizes are smaller, so nice performance win

before:
screenshot 2019-03-06 14 40 37

after:
screenshot 2019-03-06 14 40 28

@jesicarson


jesicarson:
@ascott Sorry, I know already gave you the go ahead via slack for this change, but I just did a closer review on my monitor here, and I actually think we should wait until you hear back from the Fakt foundry to make this adjustment. 
The weights are actually quite different, particularly the title of the case, and the logo. The typography was carefully considered and I don't want to lose the subtle details that the Fakt font offers. For another option, however, it looks like Inter has adjustable weights, could we try to match the original a bit more closely?

jesicarson:
homepage mocks with inter:

Screenshot 2019-03-11 22 51 33

Screenshot 2019-03-11 22 51 40

Screenshot 2019-03-11 22 52 04

ascott:
we decided to stick with Fakt.

PR #468, opened 2019-03-22
make all bookmark add/deleting work

ascott:
- generalize add/deleting bookmark handling between home/search article cards, user page article cards, and article reader views (currently only working on reader page)
- only show the bookmark icon if the user is logged in
- return 401 if user not logged in (fixed user id in controller)
- fixes: https://github.com/participedia/usersnaps/issues/325

todo: (@dethe can you help with these please? let me know if you need more info on this)
- [ ] fix the api where case, methods, orgs on home/search page always return the `bookmarked` key as false, even if it's true.
- [ ] fix the api on /users so authored things and bookmarks return the `bookmarked` key for each entry


jesicarson:
We want to show bookmark for non logged in site visitors, and direct to login/create account if clicked.

ascott:
added an issue for the backend issues https://github.com/participedia/api/issues/471

PR #481, opened 2019-04-09
fix file photo upload on edit form

ascott:
- fix bug with drag and drop functionality: previously setting the value on the file input would trigger a change event, but that is not longer working, so now we're calling the post upload event after the drop event as well. 
- sets the filename as the title attribute as a default

fixes this issue: https://github.com/participedia/usersnaps/issues/356

participedia-photo-upload

@dethe @jesicarson

### DIFF
--- a/public/css/edit-form.css
+++ b/public/css/edit-form.css
@@ -301,6 +301,7 @@ input[type=number]::-webkit-inner-spin-button {
 .edit-media-file-list[data-type="image"] img {
   display: block;
   margin-bottom: 1rem;
+  max-width: 50%;
 }
 
 .edit-multi-select-list > li,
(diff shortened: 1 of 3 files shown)


jesicarson:
Alanna, did you see the design for attribution of images display in the design guidelines? Idea is that the semi-transparent black overlay pops up on mouse-over, and if fullscreen is clicked, the full image shows on a black full screen background, with the white arrows to the left and right, and a white x in the top right corner, and the attribution text below the image. Mobile would be different, probably just the fullscreen icon and the attribution in black below the image. This is not priority, but nice to have at some point. 

Screen Shot 2019-04-09 at 3 52 31 PM

ascott:
@jesicarson i'll create an issue for this for post-launch

Issue #478, opened 2019-04-08
make sure cases/methods/orgs with location data have correct lat/long data


ascott:
currently on home page only some cases/methods/orgs have lat/long data and pins will only show on the map if it has lat/long data

<img width="953" alt="Screenshot 2019-04-08 16 28 31" src="https://user-images.githubusercontent.com/130878/55763419-a3309e00-5a1b-11e9-967a-3a1a19c3c58e.png">

PR #485, opened 2019-04-11
script to fix lat/lng for cases and organizations

ascott:
@dethe here is my first run at fixing  lat/lng for cases and organizations. 

https://github.com/participedia/usersnaps/issues/351
https://github.com/participedia/api/issues/478

for all cases and organizations
- compose an address string from the following article
  location keys: address1, city, province, postal_code, country.
- if no address string exists (meaning location keys are all falsey),
   save lat/lng as null
- if address string exists, pass that address string to geocode
  service to get lat/lng and save new lat/lng.

running this script will fix the articles in the local db, but we will still need to fix the migration fixtures (ie: https://github.com/participedia/api/blob/master/migrations/cases.json) so when the db is reset we have the correct lat/lng data. any suggestions on best way to do this? i took a stab at generating the new fixtures by pulling the data from the db once this migration was run, and then writing a new fixture json file. but then when i run the migrations, i get errors and the case table doesn't get populated. how did you generate the existing json fixtures?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -361,6 +361,15 @@
         }
       }
     },
+    "@google/maps": {
+      "version": "0.5.5",
+      "resolved": "https://registry.npmjs.org/@google/maps/-/maps-0.5.5.tgz",
+      "integrity": "sha512-RSZriyE2XViVhXgdEcQaEu3jMYh2A/jS0VahLHXqGO0VfPyEbDac4PAn7/hBJiTavWpxchKfe5OI9inJofFWxA==",
+      "dev": true,
+      "requires": {
+        "uuid": ">=2.2.1"
+      }
+    },
     "@sinonjs/commons": {
       "version": "1.4.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.4.0.tgz",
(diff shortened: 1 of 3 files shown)


dethe:
The JSON fixtures come from the old dataset, so we can't just fix it there or when we run migrations on .xyz it will go kablooey (that's the technical term). I'll need to write a migration that creates a new column for lat/long, converts the existing value into the new column, [optionally] deletes the old columns, and serves the new column as part of the JSON result.

dethe:
We are only converting existing lat/long format, not craziness like changing projections so I think we can just parse the fields and do the math, no geocode service needed, unless I'm missing something. Basically, what is done here: https://adamprescott.net/2013/07/17/convert-latitudelongitude-between-decimal-and-degreesminutesseconds-in-c/ but in SQL or one of the supported languages inside Postgres. I'll take a stab at that migration.

ascott:
@dethe since your other PR takes care of the migration, i updated this PR to only include the one-off script to geocode an address and save the lat/lng.

dethe:
@ascott OK, is this PR safe to merge then?

ascott:
@dethe :shipit:

PR #492, opened 2019-04-16
488 update authors

dethe:
Last updated now gets updated even if changes were only to metadata. Also fixed up some tests (and some bugs in code).


api/controllers/case.js
```diff
@@ -380,7 +381,7 @@ async function getCaseEditHttp(req, res) {
 async function getCaseNewHttp(req, res) {
   const params = parseGetParams(req, "case");
   params.view = "edit";
-  const article = JSON.parse(CASE_STRUCTURE);
+  const article = CASE_STRUCTURE;
```
ascott: we need to parse this as json because fs.readFileSync at https://github.com/participedia/api/pull/492/files#diff-07780e70b8cc879c37de679a75e0e6bbR33 returns a buffer and the view needs json.

dethe: Oh frack. My bad merge. I read the change backwards.

Also, we can return as a string if we pass an encoding, I forgot that nasty bit of Node weirdness.

PR #502, opened 2019-04-17
499 fix scope

### LINKED ISSUE #499: case edit: updating scope_of_influence field
i'm sending this field as `scope_of_influence: { key: 'organization' }` in the req.body, but the field value doesn't get saved. @dethe is there a different format you are expecting for this field?

dethe:
Fixes #499 

* Adds tests for all single-value lookup fields.
* Expects flattened fields (just keys) on edit.
* Work around for some values missing from DB.

This is the first step towards removing localization from the db part of the api.


ascott:
thanks @dethe, i made the needed frontend changes to with this PR.

Issue #513, opened 2019-04-24
case edit: problem saving evaluation_reports 


ascott:
when saving `evaluation_reports` i get the following error: 
i think it's an issue with the column type in postgres. it should be the same column type as photos and files. 

```
 1: { error: malformed record literal: "("https://s3.amazonaws.com/uploads.participedia.xyz/8e0c59f8-53df-47a9-8e53-92a7bae98daa-TEXT.TXT","",,TEXT.TXT)"
            at Connection.parseE (/Users/alannascott/code/participedia/api/node_modules/pg/lib/connection.js:555:11)
            at Connection.parseMessage (/Users/alannascott/code/participedia/api/node_modules/pg/lib/connection.js:380:19)
            at Socket.<anonymous> (/Users/alannascott/code/participedia/api/node_modules/pg/lib/connection.js:120:22)
            at Socket.emit (events.js:193:13)
            at Socket.EventEmitter.emit (domain.js:481:20)
            at addChunk (_stream_readable.js:296:12)
            at readableAddChunk (_stream_readable.js:277:11)
            at Socket.Readable.push (_stream_readable.js:232:10)
            at TCP.onStreamRead (internal/stream_base_commons.js:150:17)
          name: 'error',
          length: 220,
          severity: 'ERROR',
          code: '22P02',
          detail: 'Too many columns.',
          hint: undefined,
          position: '778',
          internalPosition: undefined,
          internalQuery: undefined,
          where: undefined,
          schema: undefined,
          table: undefined,
          column: undefined,
          dataType: undefined,
          constraint: undefined,
          file: 'rowtypes.c',
          line: '267',
          routine: 'record_in' 
}
```

dethe:
Fixed by #519

PR #519, opened 2019-04-26
Update shape of evaluation_reports

dethe:
Fix for #513 

You will need to run `npm run resetlocaldb`

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -196,11 +196,11 @@ function getUpdatedCase(user, params, newCase, oldCase) {
   }
 
   // media lists
-  ["links", "videos", "audio", "evaluation_links"].map(
-    key => cond(key, as.media)
-  );
+  ["links", "videos", "audio"].map(key => cond(key, as.media));
   // photos and files are slightly different from other media as they have a source url too
-  ["photos", "files", "evaluation_reports"].map(key => cond(key, as.sourcedMedia));
+  ["photos", "files", "evaluation_reports", "evaluation_links"].map(key =>
+    cond(key, as.sourcedMedia)
+  );
   // boolean (would include "published" but we don't really support it)
   ["ongoing", "staff", "volunteers"].map(key => cond(key, as.boolean));
   // yes/no (convert to boolean)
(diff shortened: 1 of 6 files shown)


api/controllers/case.js
```diff
@@ -196,11 +196,11 @@ function getUpdatedCase(user, params, newCase, oldCase) {
   }
 
   // media lists
-  ["links", "videos", "audio", "evaluation_links"].map(
-    key => cond(key, as.media)
-  );
+  ["links", "videos", "audio"].map(key => cond(key, as.media));
   // photos and files are slightly different from other media as they have a source url too
-  ["photos", "files", "evaluation_reports"].map(key => cond(key, as.sourcedMedia));
+  ["photos", "files", "evaluation_reports", "evaluation_links"].map(key =>
```
ascott: `evaluation_links` should be like links, videos, and audio. it does not handle uploaded media.

migrations/migration_031.sql
```diff
@@ -13,13 +13,21 @@ CREATE FUNCTION urls_to_links(urls text[]) RETURNS full_link[]
...
+
+ALTER TABLE cases ADD COLUMN evaluation_reports_full full_file[] DEFAULT '{}';
+ALTER TABLE cases ADD COLUMN evaluation_links_full full_file[] DEFAULT '{}';
 
-UPDATE cases SET evaluation_reports_full = urls_to_links(evaluation_reports) WHERE evaluation_reports <> '{}';
+UPDATE cases SET evaluation_reports_full = urls_to_files(evaluation_reports) WHERE evaluation_reports <> '{}';
 
 UPDATE cases SET evaluation_links_full =
-urls_to_links(evaluation_links) WHERE evaluation_links <> '{}';
+urls_to_files(evaluation_links) WHERE evaluation_links <> '{}';
```
ascott: evaluation_links does not do uploads and functions like links, videos and audio.

dethe: Yes, but it has the same fields as files, and photos. I modified the uploader code to not upload if the url is a link.

ascott: in the frontend code evaluation_links is treated the same as links, audio and videos and has the same fields as those.  Where as evaluation_reports are treated the same as files, photos. There are common fields between them, but i'm curious why we would treat evaluation_links the same as evaluation_reports on the backend...

api/controllers/case.js
dethe: You're right, of course. I managed to mis-read the spec again.

PR #521, opened 2019-04-29
Localization - english only for now

ascott:
@dethe 

- uses [i18n-node](https://github.com/mashpie/i18n-node) library to handle localization in the app
- adds phrase app yml file for the future
- wraps all strings in the app for translation
- hides language drop down in footer, until we are ready to launch additional languages
- adds [localization readme](https://github.com/participedia/api/blob/77669fa159d52e5e81687b8d592bcef663255853/localization.md)
- adds pre-commit hook to validate locale json before committing
- removes some references to static text from the db, but more still need to be removed


dethe:
I'm going to try to get Organizations ready to merge and then we can merge this before I tackle search, how does that sound?

ascott:
@dethe sounds good 👍

PR #522, opened 2019-04-29
376 save method

### LINKED ISSUE #374: Method model changes

### LINKED ISSUE #376: Method model saving

dethe:
Update model for methods and get all tests passing. Required some refactoring of cases as well, so keep those tests passing.

Fixes #376
Fixes #374


api/controllers/case.js
```diff
@@ -195,6 +133,7 @@ function getUpdatedCase(user, params, newCase, oldCase) {
     cond("hidden", as.boolean);
     cond("original_language", as.text);
     cond("post_date", as.date);
+  } else {
```
ascott: doesn't look like we need this else block

api/controllers/case.js
```diff
@@ -226,13 +164,11 @@ function getUpdatedCase(user, params, newCase, oldCase) {
     "funder"
   ].map(key => cond(key, as.text));
   // date
-  ["start_date", "end_date", "post_date"].map(key => cond(key, as.date));
+  ["start_date", "end_date"].map(key => cond(key, as.date));
```
ascott: post_date will get sent up from the client from admin users when using the ui to update/change dates/authors on articles, should we keep it here for that case?

api/helpers/data/tagkeys.json
```diff
@@ -0,0 +1,49 @@
+[
```
ascott: i have a list of all the keys for all the select fields that are being used on the front end. i wonder if we can share that same file as the source of truth for keys. https://github.com/participedia/api/blob/master/api/helpers/shared-field-options.js#L303

api/controllers/method.js
```diff
@@ -31,16 +34,6 @@ const articleText = require("../../static-text/article-text.js");
 const methodText = require("../../static-text/method-text.js");
 const sharedFieldOptions = require("../helpers/shared-field-options.js");
 
-async function getEditStaticText(params) {
```
ascott: this is function is still needed to gather the static field options to pass to the edit views. once the localization branch is merged in it will edited, but still needed.

api/controllers/method.js
```diff
@@ -133,7 +121,123 @@ async function postMethodNewHttp(req, res) {
...
+  // boolean (would include "published" but we don't really support it)
+  ["facilitator"].map(key => cond(key, as.yesno));
+  // key
+  [
+    "facetoface_online_or_both",
+    "public_spectrum",
+    "open_limited",
+    "recruitment_method",
+    "level_polarization",
+    "scope_of_influence"
```
ascott: `scope_of_influence` belongs in the array for `as.methodkeys`. looks like you have duplicated it here.

api/helpers/db.js
```diff
@@ -303,7 +290,7 @@ function media(mediaList) {
   return simpleArray((mediaList || []).map(aMedium).filter(x => !!x)); // remove nulls
 }
 
-function sourcedMedia(mediaList) {
+function sourcedMedia(mediaList, key) {
```
ascott: doesn't look like this function uses the key param.

api/helpers/db.js
```diff
@@ -29,24 +30,7 @@ try {
...
-  //   WHERE language = 'en';
-  // `)).keys;
-  dbtagkeys = (await db.one(`
-    SELECT to_json(array_agg(key)) as keys
-    FROM rotate_tags_localized('en') AS tagvalues
-    WHERE tagvalues.key <> 'language';
-  `)).keys;
-}
-initKeys().then(() => console.log("keys initialized")); // we'll need these when users submit data
+const dbtagkeys = JSON.parse(fs.readFileSync("api/helpers/data/tagkeys.json"));
```
ascott: we could use the `shared-field-options` file i'm using for the front end for this so we don't duplicate the list of tags in the code base.
```
const sharedFieldOptions = require("../helpers/shared-field-options.js");
const tagKeys = sharedFieldOptions.tags.map(tag => tag.key);
```

api/helpers/things.js
```diff
@@ -219,21 +235,25 @@ function getEditXById(type) {
...
-            ].includes(key)
+            ["tags", "if_voting"].includes(key)
           ) {
             updatedThingFields.push({
               key: as.name(key),
               value: as.strings(value)
             });
+          } else if (
+            // fields that are rich links
+            ["links", "images", "videos", "evaluation_links"].includes(key)
```
ascott: this list should include `audio` and should not include `images`.

api/helpers/things.js
```diff
@@ -219,21 +235,25 @@ function getEditXById(type) {
...
+          ) {
+            updatedThingFields.push({
+              key: as.name(key),
+              value: as.media(value)
+            });
+          } else if (["files", "evaluation_reports"]) {
```
ascott: this list should include `photos`

api/sql/method_edit_by_id.sql
```diff
@@ -1,17 +1,42 @@
...
+    level_polarization,
+    -- key lists
+    method_types,
+    number_of_participants,
+    scope_of_influence,
+    participants_interactions,
+    decision_methods,
+    if_voting,
+    -- yesno
+    facilitator
```
ascott: i think the method spreadsheet uses `facilitator` but cases uses `facilitators`, should we use the plural for consistency? it uses the same list of options.

api/controllers/case.js
dethe: Agreed.

api/helpers/data/tagkeys.json
dethe: That would be excellent.

api/controllers/method.js
dethe: Good catch, I thought I'd fixed that already. Important note: in cases `scope_of_influence` is a single key value, but in methods it is a list of key values.

api/helpers/db.js
dethe: I use it sometimes for debugging. Can remove it.

api/helpers/db.js
dethe: Yes. I will go over all of these in one go when we merge your localization branch.

api/helpers/things.js
dethe: I think only Organizations still uses this and the whole function will go away once I update Organizations (today).

api/sql/method_edit_by_id.sql
dethe: I've been trying to use singular for things that take one result and plural for things that take multiple results, but obviously still haven't resolved all of that yet, and some fields it just doesn't work for (`audios`?). Ultimately it may not be that big of a deal, but would you prefer being consistent with cases by making it `facilitators` in methods, or by making it `facilitator` in cases?

api/controllers/method.js
dethe: OK, there is no METHOD_EDIT_STATIC query for the database, so I will put this function back, just returning an empty object for now, and populate it when we merge localization branch later today.

api/controllers/case.js
dethe: It was duplicated here, it is already in the `if(user.isadmin)` block above.

api/helpers/things.js
dethe: This function will go away later today when I update organizations. Methods, organizations, and cases used to all be similar enough that I tried using one function for updating them (in `things.js`) but they have diverged sufficiently, and I think it is easier to read and maintain, that having them each have their own update function makes more sense.

api/sql/method_edit_by_id.sql
dethe: @ascott I've resolved all of the comments except for this one, and pushed the relevant changes. Once we resolve this we can merge this branch.

dethe: Actually, we can merge now, I can normalize cases in a bugfix branch later.

ascott: i'm fine with the plural or singular as long as they are the same across article types

Issue #530, opened 2019-05-06
Map Issues


ascott:
changes to map functionality
- on all tab: show all cases and orgs with black pins. show featured articles with red pins
- on cases and orgs tabs: show all cases or orgs depending on tab with black pins, show featured with red pins
- methods tab shows no pins
- if with search term, map only shows results with black pins, and shows featured articles in red pins
- use /search?selectedCategory=All&resultType=map endpoint to get map results

jesicarson:
@ascott that doesn't quite sound right. 
Featured items are always red. Cases and orgs are always black. 
Search terms removes any pin from the map that doesn't match the search.

jesicarson:
Clicking the cases tab would remove any orgs pins from the map, and vice versa. Methods tab would show no pins.

jesicarson:
or maybe your second bullet point meant that featured items remain red and this is all saying the same thing :P  Sorry if i misread that.

ascott:
i think we are saying the same thing. i updated the issue to hopefully be more clear.

jesicarson:
yepppp perfect thanks!

PR #548, opened 2019-05-08
map functionality updates

ascott:
https://github.com/participedia/api/issues/530

- on all tab: show all cases and orgs with black pins. show featured articles with red pins
- on cases and orgs tabs: show all cases or orgs depending on tab with black pins, show featured with red pins
- methods tab shows no pins
- if there is a search term, map only shows results of search with black pins, and shows featured articles in red pins
- use /search?resultType=map endpoint to get map results

in this gif you can see the initial pageload with an empty cache renders the pins a bit slowly, but all other tabs render the pins quickly because the results have been cached from the first request. i set the cache timeout for 1 hour. @dethe, @jesicarson does this seem like a reasonable time frame? should it be shorter?

particpedia-map-changes

### DIFF
--- a/public/js/map.js
+++ b/public/js/map.js
@@ -1,5 +1,7 @@
+import moment from "moment";
 import mapStyle from "./map-style.js";
 import PopOver from "./GoogleMapsPopOver.js";
+import { xhrReq } from "./utils/utils.js";
 
 const markerIcon = {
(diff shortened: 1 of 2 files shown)


jesicarson:
@ascott what happens when the cache times out?

ascott:
@jesicarson when the cache times out, it will make another request to the server for fresh results.

jesicarson:
@ascott So if someone publishes a new case, it could take up to 1 hour to show up on the map? What's the trade off for making it less than 1 hour? will it affect the user experience?

ascott:
@jesicarson yes, that's right, if someone publishes a new case, it could take up to 1 hour to show up on the map. the tradeoff for making it less than an hour is that the request for map data would be slower, about 3s. we could use the same timeout as the search index which is 5 minutes. we could also work on making the map search response smaller, currently we are downloading more data than we need for the map, so if we eliminate the data we don't need it would speed up the request. 

i would suggest we make the cache timeout 5mins for now and we can revisit performance post launch.

Issue #543, opened 2019-05-07
update edit field options values with full option values from case spreadsheet


ascott:
some drop down field values are missing the extended help text in parentheses.

Issue #542, opened 2019-05-07
case edit - "Yes" should not be an option for "Decision Methods."


ascott:
"Yes" should not be an option for "Decision Methods."

https://github.com/participedia/usersnaps/issues/361

Issue #520, opened 2019-04-29
Title of Case should be Title of Method


ascott:
<img width="1280" alt="Screenshot 2019-04-28 20 49 19" src="https://user-images.githubusercontent.com/130878/56875675-38113080-69f7-11e9-8743-70cbec1c54b9.png">

and Title of Organization

PR #554, opened 2019-05-10
handle different text values for labels, etc across cases, methods and orgs - and more

ascott:
- make tab labels medium weight
- increase padding on sortby and make zindex bug fix
- handle different text values for labels, instructional text and placeholders for each article type (cases, methods, orgs)
```
// the format for an article label key is:
const labelKey = `${article.type}_${view}_${name}_label`;
// eg: case_edit_title_label
// eg: case_view_title_label
```
- translate all field option values
--- `shared-field-options.js` now returns an array of keys for each fieldName, and the values are fetched via the i18n library
```
// the format is 
const fieldOptionKey = `name:${fieldName}-key:${fieldOptionKey}`;
const longFieldOptionKey = `name:${fieldName}-key:${fieldOptionKey}-longValue`;
// eg: "name:implementers_of_change-key:experts": "Experts",
// eg: "name:implementers_of_change-key:experts-longValue": "Experts (e.g. scientists, engineers, criminologists, doctors, lawyers)"
```
- handle showing long values for the edit templates and short values in the view
- fix some field names for methods and orgs


ascott:
fixes the following issues: 
https://github.com/participedia/api/issues/543
https://github.com/participedia/api/issues/542
https://github.com/participedia/api/issues/520
https://github.com/participedia/api/issues/547

ascott:
@dethe just found a bug on the user edit template, will fix and then merge to master.

PR #561, opened 2019-05-11
489 only pass keys

### LINKED ISSUE #489: Change API to send and receive db-keys only, not localized UI values
Currently we send `{key, lookup_key, value}` triples, where the lookup_key is only for debugging and the value key can be found in the front-end given the `key`. We receive the same and strip out everything but the `key`. So we can save work (and a little bit of bandwidth) by only sending keys or lists of keys where we now send triples or lists of triples.

dethe:
There's a lot of code in this, but mostly a lot of *removed* code. Unduplicated some query files, fewer passed variables, fewer special cases, and much fewer migrations because we don't have to build the localization tables and update them.

Fixes #489


dethe:
While I don't think it is strictly necessary to run `npm run resetlocaldb` it will give you a smaller database and should run completely without errors.

Issue #616, opened 2019-05-23
Deleting entries (for good)


scottofletcher:
@ascott @dethe @jesicarson  
I asked about this a while ago, but we've had a situation come up with one of our partners that necessitates that we create a way to either 
1. delete entries for good or 
2. re-route non-admin users that click on a link or enter the exact URL of a 'hidden' entry to a page that says 'this entry has been deleted'

'Hidden' cases are still accessible if they were entered in any of the database-search fields (eg. components, specific methods, organizing entity) or if they were hyperlinked to in the narrative body. They should only be accessible by admins.

jesicarson:
Good point and this should get implemented. Probably not super high priority though, right Scott? 

At least we can hide all the tests and messy stuff, and chances are if it was entered in error it may not be linked to too many other entries (as a method/component etc) right?

Hiding is the MVP, properly hiding is nice to have. But correct me if it’s more urgent. Thanks!

scottofletcher:
@jesicarson I'm not sure how urgent it is, I'm waiting to hear back from Julien and Mark regarding the 'situation'. I'll forward you the email. I think the person whose work was plagiarized is going to be none-too-pleased that she can still access it by clicking a link

dethe:
I can return an error if a non-admin tries to access a hidden article easily enough. Will need some front-end work to handle the error and show a nice message. The message itself should be part of our UI localizable text.

scottofletcher:
@dethe great! thank you!

ascott:
do we need a page that says 'this entry has been deleted' or can we just return a general 404 page?
@jesicarson do you have a design for a 404 page?

jesicarson:
@ascott I don't have a mockup, but would be nicer to have a page, or what about our standard modal/popup/warning format? Could that work?  And yes use the text: "This entry has been removed"

jesicarson:
@scottofletcher combatting plagiarism does feel urgent.

jesicarson:
@ascott are you ok without a mock if we just use the standard modal?

scottofletcher:
Sounds good all!

ascott:
@jesicarson and i just discussed via slack, and we decided on a general 404 page to handle requests to hidden pages as well as invalid url's. @dethe i can implement the 404 response from the server if the article is hidden as well.

scottofletcher:
[...]

PR #617, opened 2019-05-23
add 404 handling, respond with 404 if article hidden for non admins

ascott:
adds handling to respond with a 404 page for invalid urls and hidden articles (for non admins)

Screenshot 2019-05-23 12 24 12

fixes https://github.com/participedia/api/issues/616

@jesicarson @dethe @scottofletcher

### DIFF
--- a/api/helpers/things.js
+++ b/api/helpers/things.js
@@ -46,6 +46,12 @@ const fixUpURLs = function(article) {
 
 const returnByType = (res, params, article, static, user) => {
   const { returns, type, view } = params;
+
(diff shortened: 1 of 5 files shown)


scottofletcher:
can we change it to "Sorry, this page no longer exists"? that might clue people in to the fact that it was removed and not that they have the wrong URL

ascott:
@jesicarson and i decided on a implementing a general 404 page so it will be used for pages that used to exist and pages that never existed, so we want to keep the copy general.

PR #628, opened 2019-05-29
map overlay updates

ascott:
- add padding to tagline blocks
https://github.com/participedia/usersnaps/issues/708
- increase size of pins on map to match legend (i got this as close as i could, but it appears that the pin size changes depending on how the map is zoomed. i matched the size at the default zoom level)
https://github.com/participedia/usersnaps/issues/710
- adjust tagline width to account for large screens widths
https://github.com/participedia/usersnaps/issues/711

Screenshot 2019-05-29 12 47 48
Screenshot 2019-05-29 12 47 31
Screenshot 2019-05-29 12 46 57
Screenshot 2019-05-29 12 47 10

@jesicarson

### DIFF
--- a/public/css/home-search.css
+++ b/public/css/home-search.css
@@ -91,6 +91,9 @@
   background-color: #fff;
   font-size: 32px;
   line-height: 47px;
+  -webkit-box-decoration-break: clone;
+  box-decoration-break: clone;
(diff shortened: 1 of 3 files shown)


jesicarson:
@ascott nice! good to go!

PR #636, opened 2019-05-31
fix evaluation_reports bug

ascott:
some `evaluation_reports` urls that were uploaded with the old xyz ui were not saved with the aws base url. this was causing an error because our file upload code was only expecting to see either a data uri or a url that starts with http.

with this change the file upload function with check that the string is a data uri before uploading. if it's not a data uri and it's not already an aws url, add the aws base url.

fixes: https://github.com/participedia/usersnaps/issues/722

### DIFF
--- a/api/helpers/db.js
+++ b/api/helpers/db.js
@@ -398,11 +398,12 @@ function aSourcedMedia(obj) {
   }
   if (!obj.url) return null;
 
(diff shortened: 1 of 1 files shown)


ascott:
fixed by manually updating the fields in the db.

PR #637, opened 2019-05-31
autocomplete improvement

ascott:
- fix autocomplete field so that it shows a match if the value is found anywhere in the string not just at the beginning
- also made the matched text bold so it's easier to see

fixes: https://github.com/participedia/usersnaps/issues/728

participedia-autocomplete-2

@jesicarson @scottofletcher


scottofletcher:
Awesome! And I like the bold! Fancy~~~

PR #638, opened 2019-05-31
updates the "Do Full Version" flow

ascott:
Updates the "Do Full Version" flow, so we show the full version fields without saving the current changes or discarding them.

participedia-do-full-version-flow-2

fixes: https://github.com/participedia/usersnaps/issues/651

@jesicarson

### DIFF
--- a/public/js/edit-form.js
+++ b/public/js/edit-form.js
@@ -108,7 +108,7 @@ const editForm = {
     if (response.user) {
       // redirect to user profile page
       location.href = `/user/${response.user.id}`;
-    } else {
+    } else if (response.article) {
       // redirect to article reader page
       location.href = `/${response.article.type}/${response.article.id}`;
     }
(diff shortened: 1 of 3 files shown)


jesicarson:
The thing is, people might not want to publish yet, if their entry is still a draft. Can we keep it the way it was working on xyz, so that you could just load the full form by clicking do full version (without publishing first?)

ascott:
I didn't think i could make this work without saving the form data first, but i'm now realizing that I might be able to do it another way so that it doesn't get published. I'll see if I can get that to work. It's actually simpler, so not sure why I didn't think of that earlier. ^_^

jesicarson:
thanks!

ascott:
@jesicarson i updated the gif in the description. this way the changes don't get saved, but they don't get discarded either.

jesicarson:
perfect

PR #639, opened 2019-06-03
Canonicalize url

dethe:
It would be better to redirect at a lower level, but we get here from so many different domains I'm afraid of missing some. This should do the job.

I've already set participedia.nlet to force https from cloudflare, so we should be OK there.

### DIFF
--- a/app.js
+++ b/app.js
@@ -52,8 +52,8 @@ app.use(express.static("public", { index: false }));
 app.use(morgan("dev")); // request logging
 app.use(methodOverride()); // Do we actually use/need this?
 app.use(cors());
-app.use(bodyParser.json({ limit: "5mb" }));
-app.use(bodyParser.urlencoded({ limit: "5mb", extended: true }));
+app.use(bodyParser.json({ limit: "50mb" }));
+app.use(bodyParser.urlencoded({ limit: "50mb", extended: true }));
 app.use(cookieParser());
 app.use(errorhandler());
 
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -35,15 +35,25 @@ const { getUserOrCreateUser } = require("./api/helpers/user.js");
 
 const port = process.env.PORT || 3001;
 
+// canonicalize url
+app.use((req, res, next) => {
+  if (
+    process.env.NODE_ENV === "production" &&
+    req.hostname !== "participedia.net"
+  ) {
+    res.redirect("https://participedia.net" + req.originalUrl);
```
ascott: i think we need a slash at the end of "https://participedia.net", i don't think `req.originalUrl` starts with a slash.

app.js
```diff
@@ -35,15 +35,25 @@ const { getUserOrCreateUser } = require("./api/helpers/user.js");
...
+  next();
+});
 // CONFIGS
 app.use(compression());
 app.set("port", port);
 app.use(express.static("public", { index: false }));
 app.use(morgan("dev")); // request logging
 app.use(methodOverride()); // Do we actually use/need this?
 app.use(cors());
-app.use(bodyParser.json({ limit: "50mb" }));
```
ascott: this was changed to allow for multiple files to be uploaded at once. do you see an issue with keeping this at 50mb?

app.js
ascott: just double checked, it does start with a slash, so this looks good!

PR #653, opened 2019-06-07
586 faceted search

### LINKED ISSUE #586: home/search - faceted search backend
queries from the front end will send a request in the following format.
```
/?query=free+form+text+search&tags=accessibility,animal&general_issues=agriculture,arts&country=ca
```

design issue: https://github.com/participedia/api/issues/480

frontend pr: https://github.com/participedia/api/pull/531

dethe:
Fixes #586 

Known issues:

1. Countries are stored in the backend as "Canada", "United States", so we need to specify like that in queries, using `&country=ca` will not currently work.
2 We have both Germany and Deutschland, so we clearly need to do some normalization of country data.
3. We also have a country named "5000", no idea how that got in there
4. Counting search results happens before filtering, so we get some weird counts lke `1 - 1 results of 20 for "bicycles" (page 1 of 1)` when there were 20 search results for bicycle, but only one of them was in Germany. This will take longer to fix.


api/controllers/search.js
```diff
@@ -73,10 +73,11 @@ router.get("/getAllForType", async function getAllForType(req, res) {
...
 const singularLowerCase = name =>
   (name.slice(-1) === "s" ? name.slice(0, -1) : name).toLowerCase();
-// like it says on the tin
-const filterFromReq = req => {
-  const cat = singularLowerCase(req.query.selectedCategory || "Alls");
-  return cat === "all" ? "" : `AND type = '${cat}'`;
+
+// just get the type, if specified
+const typeFromReq = req => {
+  let cat = singularLowerCase(req.query.selectedCategory || "Alls");
```
ascott: looks like the fallback should be `all` here

ascott:
perhaps we should remove the country filter from the UI until we fix the issues related to this.

api/controllers/search.js
dethe: The first thing we do is strip off the last letter of Cases, Methods, etc, thus Alls

dethe:
Aside from a couple of anomolirs we're pretty consistent with country names. How hard would it be to use `Canada` vs `ca`?

dethe:
Full list from db:

```
Afghanistan
Albania
Argentina
Argentinien
Australia
Austria
Bangladesh
Belgium
Bolivia
Brazil
Bulgaria
Cameroon
Canada
Cape Verde
Chile
China
Colombia
Czech Republic
Dominican Republic
Ecuador
Egypt
Estonia
Ethiopia
Finland
France
Germany
Ghana
Greece
Guatemala
Iceland
India
Indonesia
Iraq
Ireland
Israel
Italia
Italy
Japan
Kenya
Latvia
Lebanon
Madagascar
Malawi
Malaysia
Mexico
Moldova
Mongolia
Morocco
Nepal
Netherlands
New Zealand
Nicaragua
Nigeria
Pakistan
Palestinian Territory
Peru
Philippines
Poland
Portugal
Romania
Russia
Rwanda
Somalia
South Africa
South Korea
Spain
Sweden
Switzerland
Tanzania
Thailand
Tunisia
Turkey
Uganda
United Kingdom
United States
Uruguay
Venezuela
Zimbabwe
```

dethe:
Sorry, didn't mean to close this, hit the wrong button.

We probably want to localize at least the Country part locations and use keys instead of strings, but for now I think we can get away with using the above list.

ascott:
do we want to limit search filters to only countries that already have articles? i'm thinking we probably want to include all countries in the dropdown list to filter by. this way, if someone adds a new article from a new country not already listed, it can be queried with the search filters. it may also be confusing to list only some countries in the search filter list. this is the file i'm using to populate the values for countries. i can replace the 2 character key with the full country name for now to get country filtering working.

jesicarson:
@ascott yes i think including all countries is best.

jesicarson:
@ascott on second thought Pat suggests that the limited set may be best.. But wherever you're at right now lets leave it and test it.

Issue #510, opened 2019-04-22
article edit forms: save creator and last_updated_at fields for admin


ascott:
creator and last_updated_at fields should be saved if user is admin

PR #661, opened 2019-06-26
backend for admin ui for original/updated author/date editing

ascott:
todo: 
- [ ] save the original author if that is passed up from the ui

fixes: https://github.com/participedia/api/issues/510, https://github.com/participedia/usersnaps/issues/520

### DIFF
--- a/api/helpers/db.js
+++ b/api/helpers/db.js
@@ -1,6 +1,7 @@
 const promise = require("bluebird");
 const url = require("url");
 const { isArray, isObject, isDate, isString, uniq } = require("lodash");
+const moment = require("moment");
 const options = {
   // Initialization Options
   promiseLib: promise, // use bluebird as promise library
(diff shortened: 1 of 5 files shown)


api/helpers/db.js
```diff
@@ -440,13 +441,12 @@ function yesno(value) {
...
+  if (!value) return null;
+
   if (isDate(value)) {
     return text(value);
   } else {
-    if (!value) {
-      return null;
-    }
-    return new Date(value);
+    return new Date(moment(value).format());
```
ascott: needed to use moment here because just using new Date wasn't giving us the results we would expect.
Screenshot 2019-06-26 09 31 37

dethe: Makes sense. JavaScript date parsing with the built-in date is not well defined and inconsistently implemented. If this is the only use we have for Moment, then we could avoid the extra dependency by parsing ISO 8601 ourselves, since it is designed to be easy to manipulate.

ascott:
going to merge this, but we still need the backend work to make saving the original author work. i will add a new issue for that work.

PR #664, opened 2019-06-29
handle csv download for cases, methods and orgs tabs

ascott:
makes the following urls download a csv with data for each type:
`/?selectedCategory=case&returns=csv`
`/?selectedCategory=method&returns=csv`
`/?selectedCategory=organizations&returns=csv`

@dethe i ended up reusing the `${THING}_BY_ID` sql to get all the needed localized values, but not sure if there is a more efficient way to do this.

### DIFF
--- a/api/helpers/create-csv-data-dump.js
+++ b/api/helpers/create-csv-data-dump.js
@@ -61,9 +61,7 @@ async function createCSVDataDump(type) {
     if (editedEntry.creator) {
       editedEntry.creator_id = editedEntry.creator.user_id;
       editedEntry.creator_name = sanitizeUserName(editedEntry.creator.name);
-      editedEntry.creator_profile_url = `https://participedia.net/user/${
-        editedEntry.creator.user_id
-      }`;
+      editedEntry.creator_profile_url = `https://participedia.net/user/${editedEntry.creator.user_id}`;
     }
 
     if (editedEntry.last_updated_by) {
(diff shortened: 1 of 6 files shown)


api/helpers/create-csv-data-dump.js
```diff
@@ -0,0 +1,63 @@
...
+  LIST_ARTICLES,
+  CASE_BY_ID,
+  METHOD_BY_ID,
+  ORGANIZATION_BY_ID
+} = require("./db.js");
+
+async function createCSVDataDump(type) {
+  const entries = await db.many(LIST_ARTICLES, {
+    type: type + "s",
+    lang: "en"
```
ascott: don't hardcode language here. use locale cookie and fallback to "en"

api/helpers/create-csv-data-dump.js
```diff
@@ -0,0 +1,63 @@
...
+  ];
+
+  const fullEntries = [];
+  await Promise.all(
+    entries.map(async article => {
+      const params = {
+        type: type,
+        view: "view",
+        articleid: article.id,
+        lang: "en",
```
ascott: don't hardcode language here. use locale cookie and fallback to "en"

api/helpers/create-csv-data-dump.js
ascott: upon further testing, looks like we need to hardcode "en" here until we have populated the localized_texts table with machine translations for other languages.

ascott:
@dethe does this look ok to merge to you?

api/sql/list_articles.sql
```diff
@@ -4,4 +4,5 @@ WITH all_titles AS
  ORDER BY title)
 SELECT *
 FROM all_titles
-WHERE title IS NOT NULL;
+WHERE title IS NOT NULL
+AND hidden IS NOT TRUE;
```
ascott: @dethe is it ok to only fetch articles that are not hidden here?

ascott:
@dethe made updates to this after getting feedback from scott and pat here: https://github.com/participedia/api/issues/681. it's ready for review again.

Issue #666, opened 2019-07-09
localization changes - scenario a


ascott:
- [x] update code to save original language as the language currently being viewed (@ascott)
- [x] update get_ARTICLE_TYPE_by_id sql to fallback to original_language if we don't have a translated version of the localized_texts in the users language. this will fix the bug where articles submitted as an original_language other than english do not render properly when viewing the site in english https://github.com/participedia/usersnaps/issues/773 (@dethe)
- [x] adding language drop down to entries and email request for translation button (design TBD from jesi) (@ascott)

PR #667, opened 2019-07-09
localization changes - scenario A frontend

ascott:
part of issue: https://github.com/participedia/api/issues/666

- when creating a new article save `original_language` field as the language that the site is being viewed in.


jesicarson:
Yes! I will work on design for the additional drop down placement on the reader as soon as I have my Adobe CC up and running (it just expired, waiting for Amber to green light my renewal at ECUAD). It can wait anyway, this progress is great!

PR #670, opened 2019-07-10
back fill user join_date fields

ascott:
fixes: https://github.com/participedia/usersnaps/issues/720


scripts/update-user-join-date.js
```diff
@@ -0,0 +1,29 @@
+const fs = require("fs");
+const parse = require("csv-parse");
```
ascott: don't need this package

PR #679, opened 2019-07-16
404 crashing

dethe:
This fixes a problem where we were trying to redirect after a 404 response had been sent. While running tests I also found some issues with localization for lists of article types, so fixed that, and some issues around asynch initialization (lists weren't ready when tests ran), so fixed that. I updated the Mocha and Chai dev dependencies too.

No issues for these, I don't think, but I was seeing the problem over and over in the log files.


api/controllers/case.js
```diff
@@ -68,7 +68,7 @@ async function postCaseNewHttp(req, res) {
     let title = req.body.title;
     let body = req.body.body || req.body.summary || "";
     let description = req.body.description;
-    let language = req.body.original_language || "en";
```
ascott: the `original_language` field  is sent up from the client as part of the body, and not as a param.

api/helpers/db.js
```diff
@@ -137,17 +140,34 @@ async function _refreshSearch() {
...
-  _listCases().then(() => console.log("cases cached"));
-  _listMethods().then(() => console.log("methods cached"));
-  _listOrganizations().then(() => console.log("organizations cached"));
-  _refreshSearch().then(() => console.log("search refreshed"));
-  db.none("UPDATE localizations SET keyvalues = ${keys} WHERE language='en'", {
-    keys: i18n_en
-  })
-    .then(() => console.log("i18n updated"))
-    .catch(error => console.error(error));
+const SUPPORTED_LANGUAGES = ["en", "de", "fr", "zh", "es"];
```
ascott: perhaps we should store this in a shared constants file so we can have a single source of truth for supported languages. we are also using this array of languages here: https://github.com/participedia/api/blob/master/app.js#L62

api/controllers/case.js
dethe: Good catch. I make sure to get original_language from the body now and also clarified places where we use original_language vs. language.

api/helpers/db.js
dethe: Good idea. We could maybe move the constants file into `api/helpers/data` at some point, but this is now using the same constants file as other code.

PR #715, opened 2019-08-18
edit forms: improve request timeout/server unavailable error handling for users

ascott:
creates a better user experience in cases where the server takes a long time to respond. currently if the server responds with a 503 error, the client doesn't handle it gracefully which results in the page just hanging on the publishing modal. 

related issues: 
https://github.com/participedia/usersnaps/issues/826
https://github.com/participedia/usersnaps/issues/805

this change adds a max request timeout of 10 seconds when adding or editing an article. if the response takes longer than 10s we send a 408 error back to the client and show the user the "Sorry something went wrong message." This allows the user to close the error so they can try submitting the form again.  If the server had crashed, they might be logged out, and so they will get the sorry you are logged out error, which will take them through the refresh login flow, and back to their article so they can submit again (without losing their inputted data). Here is a gif of that simulated flow:

participedia-server-time-out

we also want to solve the server issue of the server time-ing out, but this will at least improve the user experience when it does happen.

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -201,13 +201,6 @@ function getUpdatedCase(user, params, newCase, oldCase) {
 // Only changes to title, description, and/or body trigger a new author and version
 
 async function postCaseUpdateHttp(req, res) {
-  res.setTimeout(10000, () => {
-    // make the request timeout after 10 seconds and send 408 error to client
-    // this is to avoid the situation of the user having to wait a really long time
-    // (longer than 10s) with no feedback if the server is taking a long time to respond
-    return res.send(408);
-  });
-
   cache.clear();
   const params = parseGetParams(req, "case");
   const user = req.user;
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -201,7 +201,7 @@ function getUpdatedCase(user, params, newCase, oldCase) {
 // Only changes to title, description, and/or body trigger a new author and version
 
 async function postCaseUpdateHttp(req, res) {
-  cache.clear();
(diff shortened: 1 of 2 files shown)


jesicarson:
@ascott  I cant read the popups in the gif, what do they say? Thanks!

ascott:
@jesicarson  here are user feedback modals. let me know if you have any ideas of how to improve this. ideally the user won't see this flow very often once we get the underlying server issues resolved.

- user starts publishing entry
image

- request/server timeout happens on server
image

- user is logged out because server has restarted
image

jesicarson:
I think this is fine as is, but you can take or leave this feedback:
 
Only improvement I could see might be eliminating the "Sorry, something went wrong. Please try again" modal - what happens when they click ok? does it go directly to the next modal, or do they return to the form and have to click publish again? 

Why not just show the "refresh session" modal straight from the timeout error? (and also, you could change the wording from "save your changes" to "publish your changes" but that's nit picky)

ascott:
> Only improvement I could see might be eliminating the "Sorry, something went wrong. Please try again" modal - what happens when they click ok? does it go directly to the next modal, or do they return to the form and have to click publish again?

when they click ok they have to press publish again before going to the next modal, but i might be able to change that so it automatically tries to publish again. i'll try that.

ascott:
i updated the wording in the modal and changed things so if there is a timeout error, we automatically re-try 10 times while showing the publishing modal, rather than showing the "sorry, something went wrong modal". if after 10 attempts it doesn't work, we show the "sorry something went wrong" modal.

PR #741, opened 2019-09-04
updated logging

### LINKED ISSUE #740: Don’t log errors for db queries that actually have no data
Sometimes we will get a query for an id that does not exist in our database. We can return 404 pages for these without logging an error, this is pretty normal. We'll have to check the exception in our catch block, if it has the text 'No data returned from the query.' when we expect one or many results, no need to log.

dethe:
Fixes #740 

Check message from error in the catch block, if the error is just that there were no results, that isn't actually exceptional and we don't log it. Return a 404 in any case.

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -296,7 +296,10 @@ async function getCase(params, res) {
     fixUpURLs(article);
     return article;
   } catch (error) {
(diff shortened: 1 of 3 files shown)


ascott:
👍

PR #742, opened 2019-09-04
Return 404 for all NaN ids

### LINKED ISSUE #739: Check that articleid is not NaN before db query
Make sure we have a valid numeric id before any query that requires it (cases, methods, organizations, user). This is usually returned by `parseGetParams()`, but with malformed query strings that are missing the `?` we get NaN for the id. If we have NaN, return a 404 without throwing an exception.

dethe:
Fixes #739 

Check for `Number.isNaN(id)` and return a 404, bypassing database query altogether.


ascott:
👍

Issue #745, opened 2019-09-05
sentry error: invalid input syntax for integer: "'NaN'"


ascott:
we added Number.isNaN checks to article routes, but Number.isNaN("string") returns false.

<img width="184" alt="Screenshot 2019-09-05 12 46 55" src="https://user-images.githubusercontent.com/130878/64375518-f31b8d00-cfda-11e9-8dea-7eabf1a3c2bd.png">

we should instead check that the id is an integer.

<img width="1610" alt="Screenshot 2019-09-05 12 42 57" src="https://user-images.githubusercontent.com/130878/64375309-7b4d6280-cfda-11e9-9008-e0e2f7627f6f.png">

dethe:
OK, the quick fix the other day was a little too quick, have a better one now.

PR #746, opened 2019-09-06
fixes sentry error: invalid input syntax for integer: "'NaN'"

ascott:
for some reason `as.number` returns "NaN" (as a string) for urls with text instead of numbers, (eg: http://localhost:3001/case/www.sharecast.org.np)

fixes: https://github.com/participedia/api/issues/745


ascott:
closing in favor of https://github.com/participedia/api/pull/748

PR #747, opened 2019-09-06
use sentry to capture errors rather than console.error

ascott:
log errors to sentry instead of using console.error, and remove commented out console statements in article flow

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -212,21 +212,13 @@ async function postCaseUpdateHttp(req, res) {
     newCase.post_date = Date.now();
   }
 
-  // console.log(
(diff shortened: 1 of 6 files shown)


api/controllers/case.js
```diff
@@ -316,10 +306,6 @@ async function getCaseHttp(req, res) {
   returnByType(res, params, article, staticText, req.user);
 }
 
-function print(name, obj) {
```
ascott: unused function

api/helpers/things.js
```diff
@@ -114,19 +114,6 @@ Set.prototype.difference = function(setB) {
   return difference;
 };
 
-function compareItems(a, b) {
```
ascott: unused function

PR #748, opened 2019-09-06
catch NaN properly and also end response after 404

### LINKED ISSUE #745: sentry error: invalid input syntax for integer: "'NaN'"

dethe:
Fixes #745 

1. Much more robust checks for NaN (and in the right place this time!)
2. Log to console on from `logMessage` when on localhost
3. End response at 404 without trying to write further data to it.

### DIFF
--- a/api/helpers/things.js
+++ b/api/helpers/things.js
@@ -95,7 +95,7 @@ const returnByType = (res, params, article, static, user) => {
 };
 
 const parseGetParams = function(req, type) {
-  let parms = Object.assign({}, req.query, {
+  return Object.assign({}, req.query, {
     type,
     view: as.value(req.params.view || "view"),
     articleid: as.integer(req.params.thingid || req.params.articleid),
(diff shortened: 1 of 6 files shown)


api/helpers/things.js
```diff
@@ -95,14 +95,15 @@ const returnByType = (res, params, article, static, user) => {
 };
 
 const parseGetParams = function(req, type) {
-  return Object.assign({}, req.query, {
+  let parms = Object.assign({}, req.query, {
```
ascott: /s/parms/params

ascott:
1 typo, otherwise LGTM! :shipit:

api/helpers/things.js
dethe: Ah, that was supposed to be temporary, good eye.

PR #760, opened 2019-09-19
override page NaN to be page 1

### LINKED ISSUE #759: Unparseable page in URL
Symptom is `invalid input syntax for integer: "NaN"` in a search query.

We fixed this issue for getting cases, organizations, and methods (and I think for users), but missed URLs that pass bogus data for page number, so I'll address that and see if any other fields can be corrupted similarly.

dethe:
If the user passes in bogus data for a case id, we return a 404, but if there is bogus data for a page number, I just return page 1. If that seems like a bad call we can switch it to be a 404 instead.

Fixes #759 

Sentry alert this addresses: https://sentry.io/organizations/participedia-sentry-errors/issues/1158686160/?project=1524525&query=is%3Aunresolved&statsPeriod=14d

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -41,6 +41,9 @@ router.get("/getAllForType", async function getAllForType(req, res) {
   try {
     let objType = req.query.objType.toLowerCase();
     let page = Math.max(parseInt(req.query.page || 1), 1);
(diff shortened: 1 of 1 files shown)


ascott:
sgtm 👍

PR #769, opened 2019-10-16
[900]-After update, the original submission name changes to the most …

paninee:
…recent changes name (date stays the same).


ascott:
@paninee thanks for the fix! i rolled back the [previous PR](https://github.com/participedia/api/pull/764) so we'll need to bring those changes back into this branch.

PR #781, opened 2019-11-04
Search filters for Methods & Orgs

ascott:
@paninee this the PR includes with both your work and my work.


api/helpers/things.js
```diff
@@ -196,6 +196,72 @@ function setConditional(
   }
 }
 
+//get keylist and keys by methods,cases, orgs
+
+const facetKeys = (type) => {
```
ascott: @pan what do you think about replacing `facet*` with `searchFilters*` here to match the naming convention on the front-end?

Issue #789, opened 2019-11-20
[DEV] Add author transparency to entry reader views


ascott:
TODOs for @paninee 
- [x]  pass author/edit history data to view from controller for cases, methods and orgs
---- i'm thinking we will want to pass a key on the `article` object called `edit_history` which contains an array of objects representing each edit over time. each object should contain the user id, user name and timestamp.

```
article: {
  edit_history: [
    {
      user_id: 1,
      user_name: 'Scott Fletcher',
      timestamp:  '2018-03-05 22:14:46'
    },
   ...
  ]
}
```

TODOs for @ascott 
- [x] render list of edits in an accordion 

![image](https://user-images.githubusercontent.com/130878/69276334-2362d880-0b93-11ea-8ffb-1ee8856595b7.png)

ascott:
also remove author info and use Entry history as source of truth for edits

jesicarson:
What do you think about "Wiki Entry History" as the title? or "Wiki Contributor History" ... actually I like "contributor" better than "entry", as it is more clear. So even if we don't use "wiki" lets go with "Contributor History"

ascott:
i like either "Wiki Contributor History" or "Contributor History". Your call @jesicarson !

jesicarson:
blaaahhh! i dont know! ok let's go crazy and try "Wiki Contributor History" and see what people say :P

paninee:
Completed the API task

PR #793, opened 2019-11-29
Add contributor/editor history to entry pages

ascott:
fixes: https://github.com/participedia/api/issues/789

Screenshot 2019-12-01 11 17 48


api/sql/method_by_id.sql
```diff
@@ -12,6 +12,7 @@ WITH full_thing AS (
     completeness,
     first_author(${articleid}) AS creator,
     last_author(${articleid}) AS last_updated_by,
+    get_authors(${articleid}) as edit_history,
```
ascott: @paninee it looks like `get_authors` is returning a list of unique authors only, and we need this list to return all edits, not just from unique users.

Issue #788, opened 2019-11-18
[DEV] add language drop down to header


ascott:
![image](https://user-images.githubusercontent.com/130878/69065607-07720200-09ee-11ea-988f-210dee4d2376.png)

design issue: https://github.com/participedia/api/issues/778#issuecomment-554568734

jesicarson:
Thanks for creating this issue @ascott! Do you think for small mobile screens we just keep the drop down in the footer? Ie: my iPhone se screen won’t have space for the Lang drop down in addition to the login button. We could also consider adding it to the hamburger menu. Thoughts?

ascott:
@jesicarson i think that it makes sense to hide it in the header on small screens, and i think just leaving it in the footer for small screens will work well.

PR #794, opened 2019-12-02
Add language drop down header

ascott:
- add language select to header
- remove from reader page
- sets moment locale so dates are displayed in the correct language
fixes: https://github.com/participedia/api/issues/788

Screenshot 2019-12-01 18 39 55


jesicarson:
nice!

PR #824, opened 2020-01-13
Fix multiple filtering bug in https://github.com/participedia/usersnaps/issues/960

zphingphong:


api/helpers/shared-field-options.js
```diff
@@ -470,4 +470,5 @@ module.exports = {
     "partial_editing",
     "complete",
   ],
+  query: []
```
ascott: i'm not sure about this solution. this file is for storing the option keys for fields, and `query` isn't a field. i'm wondering if there is another solution here?

ascott:
moving your comment from the issue here for discussion: 
> The bug is

> All query parameter filter are in shared-field-option.js.
> Now, if there are multiple query. The filtering logic happened in `handlebars-helpers.js@filterCollections()`.

> And the error shows up because the query parameter query doesn't exist in shared-field-option.js.

> To fix this I added a new item in the array for query to handle this line in shared-field-option.js
> `if ( sharedFieldOptions[keys[i]].includes(arr[x]) ) {`

can you edit `handlebars-helpers.js@filterCollections()` to handle the case when the `query` param is present rather than adding it as a field in `shared-field-option.js`?

api/helpers/shared-field-options.js
zphingphong: @ascott thanks! how about this solution? https://github.com/participedia/api/pull/824/commits/43d3b5f4d5ff5c4d66eab032cc9dd4f601059aaf

ascott:
thanks for the change @zphingphong, i think this makes more sense! :+1:

PR #848, opened 2020-02-04
fix image rendering bug on home/search and reader pages

ascott:
## Description
- use `encodeURI` to fix urls with spaces in them

## Context/Issue Link
- https://github.com/participedia/usersnaps/issues/977
- https://github.com/participedia/usersnaps/issues/975
- https://github.com/participedia/usersnaps/issues/973

## How has this been tested? How can it be tested?
- copied an offending image url to my localdb to verify the issue and test the fix
- tested locally on the home/search page and on entry reader pages


ascott:
this fixed some of the issues, but not all of the issues.

Issue #868, opened 2020-03-04
[DEV] Translation of user generated text


ascott:
Overview doc: https://docs.google.com/document/d/121UEkZNXQyn9wVO4aw96CCuhjEmAASl9lXAChIHicDU/edit#

### User flows
**A new entry is added**
- a user adds a new entry and the original language is saved as the language they are viewing the site in (this is already done)
- [x] [pan] when it is saved, the user entered text fields (title, description, body) is machine translated in to our other languages, from this point the user entered text is forked, and different versions exist for title, body, and description for each language. 
----  `title`, `description`, `body` fields are saved in a table called `localized_texts`. this table already exists

**An entry is edited in its original language**
- no changes needed

**An entry is edited in another language**
- [x] [alanna] an alert is shown that says “This entry was originally added in {original_language}. [link]View this entry in {original_language}.”
- [x] [jesi] draft copy for info modal
---- Entry saved for 1st time it is machine translated into these languages
---- Each language is a version, but they share the same data (title description and body are different versions)
- [x] [pan] `title`, `description`, `body` fields on the entry object are pulled from the `localized_texts` table in the language being viewed, not in the original language
- [ ] [nice-to-have/future feature (alanna & pan)] if a user is editing an entry in a language that is not the original language, and the original language has a description/or body saved, add a prompt/button above respective input/textarea -- "Update this field with a machine translation from the original language". If the additional language version has some content in either description/body, this button is not shown.

**An entry is viewed in its original language**
- no changes needed

**An entry is viewed in a language that is not it’s original language**
- [x] [pan] The title, description, body fields on the entry object are pulled from the localized_texts table in the language the site is being viewed in
- [x] [alanna] an alert is shown that says “This entry was originally added in {original_language}. [link] View this entry in {original_language}.”

**Other tasks**
- [x] [pan] node script to machine translate all existing entries user generated fields (title, description, body) and save to `localized_texts` table. `localized_texts` contains versioned entries 
- [x] [pan] Make sure not to run the script twice on the whole set of data (expensive)
- [x] [alanna] spin up a new localization-staging server and database so we can work and test this project independently of our current staging server.  done: https://participedia-i18n-staging.herokuapp.com/
- [x] [pan] Think about where should google json file should be so that it isn’t exposed.
- [x] [pan] Accept language short code in the URL. This should switch the user's language. For example if the URL is https://participedia.net/case/de/5600, this should set the user's language to German, https://participedia.net/case/fr/5600 => French
[...]

PR #876, opened 2020-03-20
User Generated text localization

ascott:
This adds translation support for user generated content fields and a script to translate all existing  entries. 

issue with  more details: https://github.com/participedia/api/issues/868


scripts/generate-entry-i10n.js
```diff
@@ -0,0 +1,124 @@
...
+// getThings('method');
+// getThings('organization');
+
+// TEST PURPOSES
+getThings('case', 8);
+getThings('method', 6);
+getThings('organization', 6);
+
+function getThings(type, limit) {
+  db.any(`SELECT * FROM things WHERE type = '${type}' ORDER BY id DESC LIMIT ${limit}`)
```
ascott: we should only run translations on entries that are not hidden/soft deleted.

scripts/generate-entry-i10n.js
```diff
@@ -0,0 +1,124 @@
...
+
+    if (currentLanguages.indexOf(language.twoLetterCode) < 0) { // If language in loop not exist from currentLanguages. Add record
+      if (language.twoLetterCode !== data.language) { // Don't create language from original
+        console.log(`ThingID: ${thingid} => creating language for ${language.twoLetterCode} from ${data.language}`);
+        const item = {
+          body: '',
+          title: '',
+          description: '',
+          language: language.twoLetterCode,
+          thingid: thingid
```
ascott: let's also update the `timestamp` field so we know when each record was generated

scripts/generate-entry-i10n.js
ascott: @paninee did you see this comment?

scripts/generate-entry-i10n.js
ascott: @paninee also this one

scripts/generate-entry-i10n-test.js
```diff
@@ -0,0 +1,25 @@
+// Get google translate credentials
```
ascott, line 1: @paninee do we need this file? if not let's delete it.

scripts/generate-entry-i10n.js
```diff
@@ -0,0 +1,187 @@
...
+const { find } = require("lodash");
+const { db, pgp } = require("../api/helpers/db");
+
+const { Translate } = require("@google-cloud/translate").v2;
+const authKeys = JSON.parse(keysEnvVar);
+authKeys["key"] = process.env.GOOGLE_API_KEY;
+const translate = new Translate(authKeys);
+var dataLengthTotal = 0;
+
+// getThings('case');
```
ascott, line 20: i think we should add a couple functions in this file to help with testing and roll out, and so we don't have to edit and deploy a new file everytime we want to run a test on an entry or comment things out. what do you think of this?

```
// translate all non-hidden entries
function translateAllEntries() {
  ...
}

// translate a list of entries where entries is an array of objects [{ type: "case", id: 1 }]
function translateListOfEntries(entries) {
  entries.forEach(entry => {
    getThingById(entry.type, entry.id);
  });
}

function countCharactersForListOfEntries() {
  ...
}

function countCharactersForAllEntries() {
  ...
}

module.exports = { translateListOfEntries, translateAll, countCharactersForListOfEntries, countCharactersForAllEntries };

```

```
// cli usage

$ heroku run node  --app participedia-i18n-staging
// Running node on ⬢ participedia-i18n-staging... up, run.9213 (Hobby)
> 

> require("generate_entry_i10n.js");

> const entries = [
  { type: "case", id: 3193 },
  { type: "organization", id: 4560 }
];

> translateListOfEntries(entries);
```

scripts/generate-entry-i10n.js
```diff
@@ -0,0 +1,187 @@
...
+      }
+    }
+  }
+  console.log(`Data length total: ${dataLengthTotal}`);
+}
+
+async function saveRecord(records) {
+  const insert = pgp.helpers.insert(
+    records,
+    ["body", "title", "description", "language", "thingid"],
```
ascott, line 170: we should  add the timestamp data when we save a record (the field already exists in the table)

api/controllers/case.js
```diff
@@ -93,6 +95,13 @@ async function postCaseNewHttp(req, res) {
     });
     req.params.thingid = thing.thingid;
     await postCaseUpdateHttp(req, res);
+    let localizedData = {
+      body: body,
+      description: description,
+      language: original_language,
+      title: title
```
ascott, line 102: @paninee we need to add the timestamp data here

api/controllers/collection.js
```diff
@@ -81,6 +82,15 @@ async function postCollectionNewHttp(req, res) {
     });
     req.params.thingid = thing.thingid;
     await postCollectionUpdateHttp(req, res);
+
+    let localizedData = {
+      body: body,
+      description: description,
+      language: original_language,
+      title: title
```
ascott, line 90: @paninee add timestamp data

api/controllers/method.js
```diff
@@ -102,6 +103,13 @@ async function postMethodNewHttp(req, res) {
     });
     req.params.thingid = thing.thingid;
     await postMethodUpdateHttp(req, res);
+    let localizedData = {
+      body: body,
+      description: description,
+      language: original_language,
+      title: title
```
ascott, line 110: @paninee add timestamp data

api/controllers/organization.js
```diff
@@ -101,6 +102,13 @@ async function postOrganizationNewHttp(req, res) {
     });
     req.params.thingid = thing.thingid;
     await postOrganizationUpdateHttp(req, res);
+    let localizedData = {
+      body: body,
+      description: description,
+      language: original_language,
+      title: title
```
ascott, line 109: @paninee add timestamp data

api/helpers/things.js
```diff
@@ -331,6 +344,58 @@ const isValidURL = string => {
...
+  for (var i = 0; i < SUPPORTED_LANGUAGES.length; i++) {
+    const language = SUPPORTED_LANGUAGES[i];
+
+    if (language.twoLetterCode !== data.language) {
+      const item = {
+        body: '',
+        title: '',
+        description: '',
+        language: language.twoLetterCode,
+        thingid: thingid
```
ascott, line 358: @paninee we need to add timestamp data here

scripts/generate-entry-i10n.js
paninee: @ascott I normally run the script on my computer. Since I'm the one who would be taking care of the translation script. We won't have to re-deploy every time we want to test. 
I think this isn't a good use of time. Also, changing a code this late, right before we want to deploy isn't a good idea.
It also would likely to be run once.

ascott: @paninee i started on this and will push my progress to share with you.

ascott: here is what i am proposing. i'm really only changing the interface. https://github.com/participedia/api/pull/876/files#diff-1d6197377a3babe9525e11746ba8999d

scripts/generate-entry-i10n.js
```diff
@@ -0,0 +1,187 @@
...
+// getThingById("case", 3193);
+// getThingById("organization", 4560);
+// getThingById("method", 5079);
+// getThingById("organization", 5773);
+// getThingById("case", 386);
+// getThingById("collection", 5912);
+// -----------------------------------------------------------
+
+function getThingById(type, id) {
+  db.any(`SELECT * FROM things WHERE type = '${type}' AND id = '${id}'`)
```
ascott, line 46: @paninee we should only translate non-hidden entries

scripts/generate-entry-i10n-test.js
paninee: Done

scripts/generate-entry-i10n.js
paninee: done

scripts/generate-entry-i10n.js
paninee: done

api/helpers/things.js
paninee: done

scripts/generate-entry-i10n.js
paninee: done

PR #906, opened 2020-05-04
add vanity url for  covid initiative

ascott:
## Description
- add redirect from participedia.net/citizensvoicescovid to the google page site url for covid initiative page

TODO:
- [x] replace redirect url once i get it from jesi


jesicarson:
the site isn't published yet, but this is the url that I plan to publish: https://sites.google.com/participedia.net/citizensvoicescovid
i will let you know when it is live @ascott

Issue #922, opened 2020-05-26
[DEV] i18n - handle translation on really long entries


ascott:
- [x] research safe max limit of translation characters
- [x] implement function when we translate entries on the fly, to chunk translations requests so we don't run into limit issues. Only translate it in chunk if the data is more than 5000 characters.
- [x] Log Google API error to Sentry

PR #923, opened 2020-05-28
Handle Google translation in chunks for long entries

paninee:
https://github.com/participedia/api/issues/922


scripts/generate-entry-i18n-chunks.js
```diff
@@ -0,0 +1,222 @@
+// cli usage
```
ascott, line 1: @paninee do we need this? it looks like it is a duplicate for `scripts/generate-entry-i18n.js`

PR #927, opened 2020-06-06
Add limit and offset in map queries to match with pagination

paninee:
https://github.com/participedia/usersnaps/issues/1062

### DIFF
--- a/api/controllers/search.js
+++ b/api/controllers/search.js
@@ -124,10 +124,10 @@ const offsetFromReq = req => {
 
 const limitFromReq = req => {
   let limit = parseInt(req.query.limit || RESPONSE_LIMIT);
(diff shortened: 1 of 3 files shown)


api/controllers/search.js
```diff
@@ -124,10 +124,10 @@ const offsetFromReq = req => {
 
 const limitFromReq = req => {
   let limit = parseInt(req.query.limit || RESPONSE_LIMIT);
-  const resultType = (req.query.resultType || "").toLowerCase();
-  if (resultType === "map") {
-    limit = 0; // return all
-  }
+  // const resultType = (req.query.resultType || "").toLowerCase();
```
ascott, line 127: let's delete commented code please.

PR #929, opened 2020-06-10
Filter the map pins according to the search results, but not the pagination

paninee:
https://github.com/participedia/usersnaps/issues/1062


ascott:
looks good! 👍

Issue #896, opened 2020-04-08
[DESIGN] Improvements to the Research Page 


jesicarson:
@Mattygryan will review https://participedia.net/research and give feedback.

May want to add a section for Participedia Data (and distinguish from "data repository" section, that could be confusing)

Discussion about producing a simple diagram to give people a better sense of the data model.

Also makes sense to publish the draft codebook and CSV downloads here ASAP with some explanatory text that they are a work in progress.

jesicarson:
@Mattygryan is this sort of what you mean by visualizing the model? 
![Screen Shot 2020-04-08 at 3 02 02 PM](https://user-images.githubusercontent.com/13560530/78837944-f8461d80-79a9-11ea-9e78-5c30ca881605.png)
PS: I used Miro to quickly mock this up (collaborative white board tool - [check it out](https://miro.com/welcomeonboard/xLDn25rn8XDmYwBByEtAjpS3IxFJbsWbtVt12IOcBcyRutMImsG5cMJoq6yDkxG0))

Mattygryan:
> May want to add a section for Participedia Data (and distinguish from "data repository" section, that could be confusing)

This makes good sense!

Mattygryan:
This is super. Nice work! So many good ideas that clarify things for people coming at the site pages. I really think and hope this is going to help usability. Cheers folks 👍

jesicarson:
Change "Methodology" heading to "Participedia Methodology" (lucy usersnap, sept 2020)

Draft some honest data quality text - work in progress, constantly improving, etc (old usersnap)

add the codebook (veronica - Nov 2020)

Add published papers by team/anyone about participedia / using pp data

jesicarson:
Sender: lucy.parry@canberra.edu.au
Comment: I think we need to get a bit meta and also pose on this page the question what kinds of research can be done with PPedia data, for what purposes, and with what caveats.
https://github.com/participedia/usersnaps/issues/1142
![image](https://user-images.githubusercontent.com/13560530/100290089-4e62d380-2f2f-11eb-9088-a466f582b0b1.png)

jesicarson:
Alanna suggested adding text above the graph: https://github.com/participedia/api/issues/931

jesicarson:
**SURVEYS Update for Research page**

Per email thread with Mark, Pat, Matt, Bonny in March 2022 - Subject "RE: Questions about Participedia surveys" - I proposed the following:

[...]

PR #930, opened 2020-06-11
Update research page with real-time data downloads section

ascott:
## Description
- update research page with real-time data downloads section
- add a Graph Commons network map of data structure to new section
https://graphcommons.com/graphs/d743b2ae-3da8-43a3-821a-6bc05f6d103c?auto=true
- on large screens, open all accordions on page load

## Context/Issue Link
https://github.com/participedia/api/issues/896

## Screenshots (if it's a frontend change)
Screen Shot 2020-06-10 at 5 03 01 PM


jesicarson:
@ascott I assume the graph image redirects people to the graphcommons site? or does the graph load in the research page?

ascott:
@jesicarson the graph will load in page, and also you can click the graph commons text in the bottom left once you are viewing the graph in the page.

jesicarson:
very cool. i don't suppose the graph could be pre-loaded so we don't have to click it to load?

ascott:
@jesicarson looked at the options, and it is possible to have it load straight away 👍 

Screen Shot 2020-06-11 at 11 10 54 AM

jesicarson:
love it

jesicarson:
we may want to add a very short description about what the graph is showing. Something like...

The graph below is a visual representation of all the data fields for cases, methods and organizations entries, and how they are linked. For example, all three entry types share the "scope of influence" field, while only cases have a field for "legality."

PR #935, opened 2020-06-26
Fix duplicate case entry

paninee:
https://github.com/participedia/usersnaps/issues/1081

### DIFF
--- a/api/controllers/case.js
+++ b/api/controllers/case.js
@@ -79,7 +79,7 @@ async function postCaseNewHttp(req, res) {
     let body = req.body.body || req.body.summary || "";
     let description = req.body.description;
     let original_language = req.body.original_language || "en";
-    const errors = validateFields(req.body, 'case');
+    const errors = validateFields(req.body, "case");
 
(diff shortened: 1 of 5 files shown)


ascott:
@paninee could you elaborate on what the issue was and how you fixed it? thanks!

paninee:
@ascott The entry gets duplicated if there are invalid fields when the user enters invalid values.
This happened because, the logic we had before would create a new entry first, then validate other **optional** fields later. As mentioned on the usersnap, "Joyce, is unable to submit and says she's getting an error message when she tries to upload", I suspect that when she got an error she try to save again over and over. That's why there are a bunch of entries with valid required fields, but no optional field.

I also explained this briefly in Slack https://participedia-team.slack.com/archives/GN63JSWBS/p1593043348011300?thread_ts=1593029260.010900&cid=GN63JSWBS.

ascott:
@paninee thanks for clarifying for me @paninee. i missed that slack response so thanks for the link :) thanks for the fix!

PR #936, opened 2020-06-30
Handle GIF image upload

paninee:
https://github.com/participedia/usersnaps/issues/1007


api/helpers/handlebars-helpers.js
```diff
@@ -58,11 +58,17 @@ function getFirstLargeImageForArticle(article) {
 }
 
 function getFirstThumbnailImageForArticle(article) {
-  const url = getFirstLargeImageForArticle(article);
+  let url = getFirstLargeImageForArticle(article);
+
   if (url) {
+    let imagePath = 'thumbnail';
+    if (url.indexOf('.gif') >= 0) {
+      imagePath = 'raw';
```
ascott, line 66: could you remind me how this is working again? this is to handle legacy gifs that were uploaded in the past, since we are no longer supporting gif uploads, is that correct?

ascott: i think it could be helpful to add a comment here about why we have a different path for gifs, otherwise this looks good to me 👍

paninee: yea, this is to handle the existing GIF.

PR #937, opened 2020-06-30
Add x icon to righ hand side of search input if there is text in the field to clear query params

paninee:
https://github.com/participedia/usersnaps/issues/1076


ascott:
Looks great, thanks @paninee!

PR #943, opened 2020-07-15
Add map marker clustering to map on homepage

ascott:
- adds clustering functionality to map markers for entries that are not featured
- render featured markers separately so that they do not get clustered and are therefore more visible

Screen Shot 2020-07-14 at 9 51 41 PM

participedia-map-clustering


jesicarson:
Nice! Leaving the featured pins out of the clusters is a great idea!
Let's run this by @amberfj during the design review today at 1pm. Thanks @ascott !

PR #945, opened 2020-07-15
use same api key env var that we are using for other GCP services for consistency

ascott:


ascott:
had to merge this branch in manually, because github wasn't updating the merge ui on this PR

PR #950, opened 2020-07-24
Improve map markers and map style

ascott:
- replace marker cluster icons with the location pin icon style
- update map style to add a bit more colour while keeping it fairly simple/subdued

Before:
Screen Shot 2020-07-23 at 10 16 21 PM
Screen Shot 2020-07-23 at 10 16 13 PM

After:
Screen Shot 2020-07-23 at 10 14 46 PM
Screen Shot 2020-07-23 at 10 13 58 PM

@jesicarson


jesicarson:
Thanks @ascott! The pins are looking good, and I think we should get @amberfj to have a look at this before launching since the change is based on her feedback. She will probably want to see a mockup or screenshot of the full screen to see what the new design looks like with the results below it, to get the full picture. 

Does the blue you've selected tie in with any of the new colours being introduced on the new homepage design, such as the Humaaans illustrations?  

I'm wondering if we should take the time to do a colour review, and build a palette for the site that the map would work with also. 

One option would be to just adjust the pins for now on prod, and deal with colour as a larger project later. But let's see what Amber thinks.

public/js/map.js
```diff
@@ -223,25 +223,37 @@ const map = {
...
+          url: "/images/map-marker-images/m5.svg",
           width: 50,
-          height: 50,
-          className: "custom-clustericon-3",
-        },
+          height: 75,
+          anchorText: [-10, 0],
+          anchorIcon: [75, 25],
+          textSize: 14,
+          textColor: "#fff",
```
paninee, line 251: @ascott Should we make this `textColor` a variable? They're all duplicates of white. In case we need to change it in the future, we can just change it in a single place.

ascott:
@jesicarson good points about changing the map colour. i'll remove the map colour changes and we can discuss those later, so we can get these marker changes out now.

here's how it's looking with original map colors

Screen Shot 2020-07-28 at 9 09 58 AM
Screen Shot 2020-07-28 at 9 09 50 AM

jesicarson:
Cool thanks @ascott . This looks pretty good I think. I am wondering if the small, individual pins should have the circle in them, like they do now? Or do you prefer the solid pins?

jesicarson:
Plus it would be good if @amberfj could have a peek at this before we launch.

ascott:
@jesicarson 
> I am wondering if the small, individual pins should have the circle in them, like they do now? Or do you prefer the solid pins?
since we have the numbers in the larger pins, i found that the white circle in the individual pins looked a bit confusing? like is it a 0? so that is why i made them solid. what do you think?

jesicarson:
Yep that makes sense. Didn't think of the zero interpretation.

PR #956, opened 2020-08-05
Get language filter from getLanguage(req)

paninee:
https://github.com/participedia/usersnaps/issues/1091


api/controllers/search.js
```diff
@@ -34,8 +34,7 @@ function getLanguage(req) {
   // we can use the locale cookie to query by language.
   // currently if the locale is set to something other than "en", no results are returned,
   // so hardcoding "en" here
```
ascott, line 36: would be great to update this comment to reflect the change as well.

PR #962, opened 2020-08-14
Adjust map marker styles for smaller screens

ascott:
## Description

Adjust map marker sizes for the 2 largest sizes so they are not quite as big on mobile. 

fixes: https://github.com/participedia/usersnaps/issues/1109

before:
Screen Shot 2020-08-14 at 11 55 44 AM

Screen Shot 2020-08-14 at 11 55 36 AM

after: 
Screen Shot 2020-08-13 at 10 28 30 PM

Screen Shot 2020-08-13 at 10 28 39 PM


jesicarson:
Looks good here. I have the smallest iPhone so I can test once this is on prod. Or is it on staging? @ascott

ascott:
@jesicarson this isn't on staging yet, but i will ship it now so you can check it out.

ascott:
@jesicarson this is on prod now

jesicarson:
Thanks! This seems better. I have a hard time with zooming on mobile, but we can address this after the homepage relaunch, since the map will not be as prominent.

PR #979, opened 2020-09-04
add inspectlet code

ascott:
## Description

Adds inspectlet code for user research/analysis
We should exclude our own visits to the site by using this link: 
https://www.inspectlet.com/exclude/929677093


jesicarson:
done, I'm excluded! thanks @ascott 
looking forward to seeing some data from this

Issue #983, opened 2020-09-16
[DEV] updates to CSV data downloads


ascott:
These are changes needed to improve the format and quality of the CSV download files based on the recommendations from Verónica Hurtado in [this report](https://docs.google.com/document/d/1iFO88fUJ_P2iRTrlyRBoepIkpeabl1sPj0_mg0zlQu4/edit) and from discussion in the Sept 15 Data Management meeting [[agenda]](https://docs.google.com/document/d/11xvgrBG2l95m7CZITHaw7-R9fkWZ0xFJcuI-5HChmqY/edit).

#### Changes for CSV generator script

- [x] change date format for all date fields from ISO to YYYY-MM-DD
- [x] create multiple columns for all multi select fields
    - currently multi selects are stored in a single column as an array, ie: [term, term, term]
    - the change is to make a column for each value in the multi select, so each value in the array should have it’s own column, like purpose_1, purpose_2, purpose_3, purpose_4, purpose_5. (to a max of 5)
- [x] change all binary variables to use 0 and 1
    -  we have variables that report “TRUE” or “FALSE”, and others that report “yes” or “no”. homogenize these values as 0 and 1 across all fields like this.
- [x] include collections as a new column

#### Changes in DB/Webapp

- [x] check what max count is on all multi selects
    - are there any over 5? 
    - what should we do with options over 5? delete them?
- [x] limit all multi selects to max 5 if there is not limit on it yet
- [x] duplicate values
    - In scope_of_influence, we see that the original coding strategy generates duplicate values. In this case, we have information for “metropolitan_area” and “metropolitan” and we count them as separate categories. I suggest we make sure that the values matches so we are not missreporting some of these values.
    - we need to generate a list of all fields that contain duplicate values

paninee:
Sample CSV

[participedia-data-cases.zip](https://github.com/participedia/api/files/5343548/participedia-data-cases.zip)

paninee:
Cleaned up following duplicate values in DB ased on api/helpers/shared-field-options.js

**scope_of_influence**
citytown => city/town
metropolitan_area => metropolitan
neighborhood => neighbourhood

ascott:
@jesicarson  @paninee here is a spreadsheet documenting the old db values that are still in our DB and need to be replaced with the equivalent new values. https://docs.google.com/spreadsheets/d/12my0Il-4vFRcFefqJzE4OPUZAMT7lB5FtTxJYuQLaVc/edit#gid=740848077.

[...]

PR #986, opened 2020-10-14
Updates to CSV data

paninee:
https://github.com/participedia/api/issues/983

### DIFF
--- a/api/helpers/create-csv-data-dump.js
+++ b/api/helpers/create-csv-data-dump.js
@@ -264,7 +264,7 @@ async function createCSVDataDump(type, results = []) {
     // make sure all date fields are in the same format, ISO 8601
     const dateFields = ["post_date", "updated_date", "start_date", "end_date"];
     dateFields.forEach(field => {
-      editedEntry[field] = moment(editedEntry[field]).toISOString();
+      editedEntry[field] = moment(editedEntry[field]).isValid() ? moment(editedEntry[field]).format("YYYY-MM-DD") : "";
     });
 
     // convert primary_organizer and is_component_of into three new fields (id, title, url)
--- a/api/helpers/create-csv-data-dump.js
+++ b/api/helpers/create-csv-data-dump.js
@@ -93,6 +93,7 @@ const orderedCaseFields = [
   "audio_count",
   "evaluation_reports_count",
   "evaluation_links_count",
+  "collections"
 ];
 
 const orderedMethodFields = [
@@ -131,6 +132,7 @@ const orderedMethodFields = [
   "videos_count",
   "links_count",
   "audio_count",
+  "collections"
 ];
 
 const orderedOrganizationFields = [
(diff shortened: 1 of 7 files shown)


api/helpers/create-csv-data-dump.js
```diff
@@ -131,6 +161,7 @@ const orderedMethodFields = [
   "videos_count",
   "links_count",
   "audio_count",
+  "collections"
```
ascott, line 164: we don't need to include collections in the csv. let's remove this.

api/helpers/create-csv-data-dump.js
```diff
@@ -170,6 +201,32 @@ const orderedOrganizationFields = [
...
```
ascott, line 204: we are not including collections in the csv

views/case-edit.html
```diff
@@ -62,9 +62,9 @@ <h2 class="fieldset-header">{{t "process_sectionlabel"}}</h2>
     {{> article-edit-select name="facilitator_training" }}
     {{> article-edit-select name="facetoface_online_or_both" }}
     {{> edit-multi-select ranked=true name="participants_interactions" max=3 }}
-    {{> edit-multi-select name="learning_resources" }}
-    {{> edit-multi-select name="decision_methods" }}
-    {{> edit-multi-select name="if_voting" }}
+    {{> edit-multi-select name="learning_resources" max=5 }}
```
ascott, line 65: rather than setting a max inline here, let's set the default max to 5 in `edit-multi-select.js`.

something like this in https://github.com/participedia/api/blob/homepage-2020/public/js/edit-multi-select.js#L113 : 
```
const DEFAULT_MAX_ITEMS = 5;
const maxItems = currentList.getAttribute("data-max") || DEFAULT_MAX_ITEMS;


```

views/partials/edit-autocomplete.html
```diff
@@ -3,6 +3,7 @@
   type="text"
   class="js-edit-autocomplete"
   data-name="{{name}}"
+  data-limit="{{limit}}"
```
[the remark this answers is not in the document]
ascott: ah i see, it is for other fields as well like `specific_methods_tools_techniques`.

views/partials/edit-form.html
```diff
@@ -33,7 +33,7 @@ <h2 class="fieldset-header">{{t "Admin Only"}}</h2>
               {{> edit-boolean name="hidden" quick-submit=true}}
               {{# if (isNotCollection article)}}
                 {{> article-edit-select name="completeness" quick-submit=true}}
-                {{> edit-autocomplete name="collections" multi=true }}
+                {{> edit-autocomplete name="collections" limit=0 multi=true }}
```
ascott, line 36: let's use `max` instead of `limit` like edit-multi-select for consistency

public/js/edit-autocomplete.js
```diff
@@ -68,6 +69,21 @@ const editAutocomplete = {
       currentLiEl.querySelector("input").value = item.value;
       currentLiEl.style.display = "flex";
     } else {
+      // TODO: Validat limit
```
ascott, line 72: was there an issue with this validation? would be great to include this as well.

Issue #1006, opened 2020-11-18
[DEV] Implement CSV downloads for collections


jesicarson:
On collections landing pages, add download csv button on the cases/methods/orgs tabs, allowing downloading the full collection as separate csvs for each type. Exactly the same as how we deal with downloading results on the search page. 

![Screen Shot 2020-11-18 at 2 22 33 PM](https://user-images.githubusercontent.com/130878/99595477-a33fa080-29a9-11eb-84c6-7ae59bd94d1c.png)

paninee:
Merged into `master`

ascott:
this is on staging now, looks like there is bug where the csv does not include any records. also introduced a regression on the main search csv download

paninee:
Live and the CSV includes correct records now.

plscully:
> @paninee  -- Awesome! Thank you!
> > Live and the CSV includes correct records now.

PR #1015, opened 2020-11-27
Implement CSV downloads for collections

paninee:
https://github.com/participedia/api/issues/1006

### DIFF
--- a/public/js/csv-generator.js
+++ b/public/js/csv-generator.js
@@ -1,0 +1,5 @@
+const csvGenerator = {
+
+}
+
+export default csvGenerator;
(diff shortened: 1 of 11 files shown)


jesicarson:
I'm being nosey, and this is probably in progress, but I don't see the Results CSV button on collections pages yet.

public/js/csv-generator.js
```diff
@@ -0,0 +1,13 @@
+const csvGenerator = {
```
ascott, line 1: let's fix the white space on this file please.

public/js/csv-generator.js
```diff
@@ -0,0 +1,13 @@
...
+	init() {
```
ascott, line 2: i think this module name could be more clear, what do you think of `downloadCsvButton.js` instead

ascott:
@paninee i just left a couple comments. when you have reviewed/updated, please ship and test on staging and then to prod. thanks!

